PR #266, opened 2014-09-09
Feature/double rainbow

### LINKED ISSUE #262: First I thought it was fun. But then I liked the coloured version so much it was incredible. OMG.
Make this ( http://postimg.org/image/rs85twhff/ ) a command line option.

Debilski:
Closes #262.


Debilski:
Should be kept as an easter egg but possibly hidden behind an env variable. We don’t want to clutter argparse and the command line API with this.

Issue #269, opened 2014-09-13
Crash on timeout


khrapovs:
git[master] $ ./pelitagame --standalone-client ../group4/ tcp://172.21.3.110:2002
Using factory '../group4/' -> 'Stas and Chris Team' (standalone)
Exception in client code for team SimpleTeam('Stas and Chris Team', <group4.chris_player.ChrisPlayer object at 0x7fb69feb01d0>, <group4.chris_player.ChrisPlayer object at 0x7fb69feb0290>).
Traceback (most recent call last):
  File "./pelitagame", line 652, in <module>
    run_game()
  File "./pelitagame", line 514, in run_game
    sys.exit(client.run())
  File "/home/khrapov/Dropbox/Computation/Python/ASPP2014/pelita/pelita/simplesetup.py", line 509, in run
    self._loop()
  File "/home/khrapov/Dropbox/Computation/Python/ASPP2014/pelita/pelita/simplesetup.py", line 530, in _loop
    retval = getattr(self, action)(*_data)
  File "/home/khrapov/Dropbox/Computation/Python/ASPP2014/pelita/pelita/simplesetup.py", line 551, in get_move
    return self.team.get_move(_args, **kwargs)
  File "/home/khrapov/Dropbox/Computation/Python/ASPP2014/pelita/pelita/player.py", line 74, in get_move
    return self._bot_players[bot_id]._get_move(universe, game_state)
KeyError: 1

Debilski:
Explanation of the issue:

peilta runs a remote game to a tcp player that has not connected yet and sends it the `set_initial` and `get_move` instructions. Eventually, we run our pelita-player over tcp and pelita sends us the next `get_move`. Since our code is not properly initialised (no `set_initial`), our player crashes.

Option a) Fix `pelita` to kick out players that did not respond to the `set_initial` phase.
Option b) Fix player code to be able to play even without a proper `set_initial` phase. (This will probably lead to broken client code since we don’t start in round 0…)

Debilski:
Closing. I think we don’t support the specific network mode anymore but in any case #725 should hopefully fix a similar situation by just letting a player fail when set_initial receives a timeout.

PR #271, opened 2014-09-16
BF: Aborting with better error message when set_initial was not called before get_move.

Debilski:
Problem: A client connects while the game is already running and therefore never receives a set_initial request. Then get_move is called and the team is in an undefined state.

A better solution would be for the server to disallow a late connection in the first place.

Gives a better error message to #269.


Debilski:
TODO: Needs at least a test case attached.

Debilski:
This might have been the problem that @djarecka experienced on her MacBook as well.

PR #276, opened 2015-03-22
BLD: Speedup travis runs by not needing sudo rights.

Debilski:
This makes travis use a container instead of a fuller VM.


keszybz:
It's strange that this is not the default...

stefanv:
This is a fairly new feature.

Issue #279, opened 2015-05-18
Replace `time()` in timeout checks with `monotonic()` on Python 3


Debilski:
This is done wrong in loads of projects. We should provide a good example of doing it correctly.

stefanv:
Rally the troops!

... and thanks, I learnt something new :)

Debilski:
Only problem is the this means littering the code with things like

    time_start = time.monotonic() if six.PY3 else time.time()

and then I’m not sure anymore this is really worth implementing. :)

stefanv:
Why not just implement a utility function somewhere?

Debilski:
Guess that’s slightly over-engineered:

    @contextmanager
    def time_monochronic():
        ts = time.monotonic() if six.PY3 else time.time()
        yield lambda: (time.monotonic() if six.PY3 else time.time()) - ts

    with time_monochronic() as tm:
        print(tm())
        time.sleep(1)
        print(tm())
        print(tm())

PR #280, opened 2015-05-18
BF/ENH: Use monotonic time for time differences when possible

Debilski:
Yeah, I know. It looks rather ugly. We should drop Python 2. :)


Debilski:
Closes #279

Issue #281, opened 2015-05-22
[discussion] Syntax: Food in maze[pos]


Debilski:
Should we get rid of the syntax

    Food in maze[pos]

Since Python offers no first-class support for singleton symbols[*], we currently have `Food`, `Free`, and `Wall` defined as single-character strings and `maze[pos].__contains__` is overloaded to return a multi-character string. While the syntax looks quite nice, I do not really like its underlying representation. Additionally, since it is not possible to query

     Bot(1) in maze[pos]

it is unclear, if special casing the food (which does not even differentiate between enemy and own food) is really worth it.

I don’t have the students’ repositories with me right now, so I cannot make any guesses how often this syntax has been used (in the tutors’ repo it is only used by Pietro and me in three lines altogether) but in many cases, I’d think the stateful helper methods on `Player` provide enough information for the neighbourhood already and for global scales, we’d probably resort to boolean matrices `maze.food[pos]`. (Exact procedure might vary depending on internal optimisations regarding the data structure.)

Is there anyone who especially likes the old syntax?

[*]: Alternatives would be classes, which are weird because they could be initialised (which would be senseless) or objects, which would not be singletons and we’d have to do instance checks.

keszybz:
On Fri, May 22, 2015 at 07:15:12AM -0700, Rike-Benjamin Schuppner wrote:
> Should we get rid of the syntax
> 
>     Food in maze[pos]
> 
> Since Python offers no first-class support for singleton symbols[*],
There's enum in Python 3, backported as a module to Python 2, iirc.
Is this what you mean by singleton symbols?

> we currently have `Food`, `Free`, and `Wall` defined as single-character strings and `maze[pos].__contains__` is overloaded to return a multi-character string. While the syntax looks quite nice, I do not really like its underlying representation. Additionally, since it is not possible to query
> 
>      Bot(1) in maze[pos]
> 
> it is unclear, if special casing the food (which does not even differentiate between enemy and own food) is really worth it.
> 
> I don’t have the students’ repositories with me right now, so I cannot make any guesses how often this syntax has been used (in the tutors’ repo it is only used by Pietro and me in three lines altogether) but in many cases, I’d think the stateful helper methods on `Player` provide enough information for the neighbourhood already and for global scales, we’d probably resort to boolean matrices `maze.food[pos]`. (Exact procedure might vary depending on internal optimisations regarding the data structure.)
> 
> Is there anyone who especially likes the old syntax?
[...]

Issue #272, opened 2014-11-21
Bug in pelitagame when both teams use the same module name


Debilski:
pelitagame only uses the first module for both players when both module names are equal.

Debilski:
This is a really unfortunate bug. We could avoid it by running all clients as separate processes (which would mean that the player would have to wrap the user’s code complete with command line arguments for ports). This means of course more API/protocol stuff but it also means that Python players are no longer special. They’d function exactly as players written in other languages. Thoughts: @otizonaizit

venthur:
I suggested this feature years ago. Again, I suggest to look at http://aichallenge.org/ and the way they implemented the IPC. It is basically round-based and uses STDIN and STDOUT. This allows also for programming language independent players. Have a look at it, it is really not that complicated.

Debilski:
I know, @venthur . The question is whether the error messages are then good enough whenever something goes wrong. When we run players through pelitagame, we can at least tell whether something went wrong during, say, the import stage of a user’s player. When it is just command line, errors may become really opaque.

otizonaizit:
I think running the players as separate processes seems the cleaner solution (but still you'll have to disambiguate the team names in the viewer, wouldn't you?). Reworking the whole communication protocol is a tangent issue. I personally don't see what the advantage would be, but the discussion should happen on a different issue or on the mailing list...

Debilski:
The team name gets set through the zmq API (or in Python code through the `Player` API, resp.) so this would have to be configured by the user anyway.
It is not the whole protocol that would have to be redone; mostly some initial confirmation checks would be useful and some timeout handling. This of course introduces some delay when things go wrong.

The most important part would be a command line spec. Ie.

    pelitagame ./my_player.py:factory

will have to result in a call to

    ./my_player.py --factory factory --zmqconn tcp://127.0.0.1:54321

or similar. And ideally this should also work on Windows…

Debilski:
Removing module loading related stuff from `pelitagame` and putting it into a separate script `pelita_module.py` which is then executed in a subprocess by `pelitagame` seems to be working without too much command line hassle. (There is a risk of diverging interfaces.) It may even give us the debug mode for free.

I’ll keep looking into whether a zmq free `Popen.communicate` solution gives us some advantages.

PR #283, opened 2015-05-23
Backend simplification.

Debilski:
First baby step in refactoring the backend structure.

1) `Team` does not need to know about the name (this is only relevant for game master and viewers) and also does not need to hold bot ids (these can be retrieved from the universe).

Please do not yet merge. There is more to come.


Debilski:
2) `GameMaster` loses its stateful method `register_team`. All teams must now be ready when `GameMaster` is initialised.
3) Additionally, `set_initial` is now supposed to return the team name. (TODO: Write tests for this.)

Debilski:
4) A fix for #272 is implemented in ‘ENH: Add an external program to load the players’ which also makes the whole `--standalone` mess superfluous. Now all modules are being run as a subprocess. Multiprocessing (or -threading) is not used anymore. All players have therefore (at least currently) full IO and can accept input from the command line. → debug mode is possible without further command line args (well, except for maybe timeout).

Debilski:
5) We can now easily run Python 2 modules together with Python 3 players:

    $ python pelitagame py2@../2014/old_style p3@modern_player

This runs pelitagame with whatever `python` points to (and opens a Tk window with the same executable) and runs `../2014/old_style` with Python 2 and `modern_player` with Python 3.

The code is a bit drafted and may be made more dynamic to allow for different languages/modules being added. (Think IPython kernel spec files.)

Debilski:
6) There is now a pelita/libpelita.py which will eventually contain most of the code from pelitagame and can be called from the tournament script.

Debilski:
I wanted to get rid of `Team` completely (because it is pretty redundant) but it seems harder than I thought. Leave that discussion for later. I think I want to merge now.

Debilski:
7) Food is not stored in the maze anymore (https://github.com/ASPP/pelita/issues/281)

Debilski:
8) The main process does not import the player modules anymore but runs `module_player.py` in a subprocess which takes care of that. Fixes the nasty bug of #272.

Debilski:
By popular request: Bullet points 4, 5, 6, 8 are now in https://github.com/ASPP/pelita/pull/283 .

Issue #201, opened 2012-09-06
group template and git repositories


otizonaizit:
We should remove the groupN-template.zip. We should configure the group repositories such that the first commit contains the template. This would make life much easier for git beginners during the programming project.

Debilski:
Closed in https://github.com/ASPP/pelita/pull/297

Issue #144, opened 2012-08-06
Move groupN-template.zip code to extra repository


Debilski:
Extract https://github.com/ASPP/pelita/blob/master/doc/source/groupN-template.zip and put its contents into a dedicated repository.

Cons:
— There is no indication which version of the template belongs to which version of pelita.

Pros:
— Nobody has ever looked into the zip file since last year’s contest. So we do not know if it is valid anyway. A repo would be more transparent.
— People may clone and immediately use it. (Or use github’s downloader.)
— Could use tags to indicate version…

PR #297, opened 2015-07-28
DOC/BLD: Remove zipped repo.

Debilski:
Closes https://github.com/ASPP/pelita/issues/144 and https://github.com/ASPP/pelita/issues/201


Debilski:
Closes #201

PR #298, opened 2015-07-29
ENH: Creating mazes with Python 3.

Debilski:
@otizonaizit @pberkes I have removed the `mdp.graph` import because I could not install it with pip. :disappointed:

I don’t mind keeping it in as a historic reference, though. :)


pberkes:
As long as it works...

Debilski:
Not sure if it works – or if it is ever going to be used again given we have all the layouts ready…

PR #304, opened 2015-08-20
Feature/modularisation

Debilski:
Split pull request. Going to be rebased onto master once #283 is merged.

PR #305, opened 2015-08-24
Flexible tournament

Debilski:
New, more flexible tournament mode. To be rebased when #304 is merged.

Configuration is done in a yaml file.

    location: Munich
    date: 2015
    team_prefix: group
    seed: 42
    bonusmatch: True
    teams:
      - spec: FoodEatingPlayer
        members:
          - "Player One"
          - "Player Two"
      - spec: StoppingPlayer
        members:
          - "b"
      - spec: random
        members:
          - noone

`spec` is the same module specifier that the pelitagame command line uses. So, things like `py3@/home/user/pelita_player:dumb_factory` are also possible. The name of the team will be queried during initialisation of the tournament (which also serves as a short sanity check).

Still TODO:
  - [x] Demo mode
  - [ ] More tests
  - [ ] Test integration
  - [x] KO mode handling (bonus matches)
  - [ ] Automatic state savings in case of crashes.

### DIFF
--- a/pelita/libpelita.py
+++ b/pelita/libpelita.py
@@ -10,7 +10,8 @@ import sys
 
 import zmq
 
-from .simplesetup import RemoteTeamPlayer
+from . import layout
+from .simplesetup import RemoteTeamPlayer, SimpleController, SimplePublisher, SimpleServer
 
 _logger = logging.getLogger("pelita.libpelita")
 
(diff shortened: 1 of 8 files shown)


Debilski:
I’d like to merge with all the current bugs so that @otizonaizit @pberkes and @keszybz can test it. :)

Debilski:
Merging now so that we have mass exposure for bug tracking.

PR #311, opened 2015-09-02
fixed name and year of school

jakobj:


Debilski:
Closes https://github.com/ASPP/pelita/issues/254

PR #327, opened 2015-09-10
Feature/testing for tournaments

Debilski:
WIP


Debilski:
Fails on 3.3 and 3.4 because of gratuitous Python 3.5 syntax.

Debilski:
Branch is rather stable now. Going to add a README for the tournament and do some more cleanup.

Debilski:
Interactive mode and spoken output might have some bugs. Other than that it can be tested.

otizonaizit:
Should the tournament not be run in interactive mode by default? That would make sure that we don't make the mistake of running it non-interactively at the school. Even if we have a recovery mode now, I still think the interactive mode should be default. Non interactive is only useful during testing, right?

Debilski:
Maybe. OTOH, one notices rather quickly that the non-interactive mode is on. But you’re correct I guess. Still, I would disable the sound output by default.

otizonaizit:
Or maybe, create a combined option `--live` which would activate 
whatever we deem correct for the live tournament. It could for example 
be equivalent to interaactive, speaking, 300 rounds. What else?

Debilski:
The main reason I removed the default speaking was because I also removed the checks whether the binary was found and the like. It is needed once a year and not really worth cluttering so much of the setup code with if statements.

OTOH I agree that there could be a config shortcut for a bunch of settings (and a more well-thought out configuration system in general).

otizonaizit:
Yeah, but the tournament **is** indeed needed once a year, and that once needs to work out of the box and tell you what is wrong when something is wrong and fail gracefully when nothing better can be done. That's the whole point of automatize it. So, I wouldn't describe the checks as clutter...

Debilski:
Yeah, okay. The problem is that it has gotten harder to test for the *proper* call anyway since we are calling `shlex.split` on whatever string is passed to `--speaker`. I can test that but if the parameters are wrong, we’re out of luck anyway.

Maybe we need to have an interactive configuration script with questions ‘Did you hear a sound?’, ‘How many rounds?’, ‘Where are the teams?’ that gives us back a `tournament-defaults.yaml` to be on the safe side.

otizonaizit:
> Maybe we need to have an interactive configuration script with
> questions ‘Did you hear a sound?’, ‘How many rounds?’, ‘Where are the
> teams?’ that gives us back a `tournament-defaults.yaml` to be on the
> safe side.

That is indeed a good idea. What about `tournament --setup`?

Debilski:
Fine with me. I’ll think about something.

Debilski:
It’s not beautiful but it is something.

Debilski:
Fixing up the logging later. Otherwise, I’ll assume that @jakobj did the review and accepted, right?

PR #332, opened 2016-03-25
Feature/python 3

### LINKED ISSUE #323: Drop support for Python 2
There is no hurry since `six` is fine for most of our purposes. But no-one missed it during the last school and if we drop it, we could probably immediately go to Python 3.4 as a baseline. (This means Debian wheezy and Ubuntu 12.04 – and OS X – need to use something like anaconda.)

Debilski:
Work-in-progress branch. Closes #323.

We could also get rid off the utf-8 header declarations and I’ll also go through the rest of the code for Python 2 traces before merging.


keszybz:
+1

Debilski:
Any obvious oversights? Then I’m merging.

keszybz:
On Thu, Mar 31, 2016 at 07:06:22AM -0700, Rike-Benjamin Schuppner wrote:
> Any obvious oversights? Then I’m merging.
+1

PR #334, opened 2016-04-14
Feature/ui rotation

Debilski:
Tries to fix ui bugs but has the potential of introducing runtime type errors. There should probably be a better way.


Debilski:
Merging. Please report all failures immediately!

PR #361, opened 2016-09-06
BLD: Use codecov.io for coverage.

Debilski:


Debilski:
Let time decide, what is better…

PR #364, opened 2016-09-07
NF: Shuffle teams in a deathmatch.

### LINKED ISSUE #353: home/away team selection needs to be shuffled in death matches
Currently the best team of round 1 will always be blue.

Debilski:
This shuffles the order of teams in a deathmatch. Unfortunately,
this is not stored in the state because shuffling is only done inside the
start_match function.

Closes #353


Debilski:
Needs review. @otizonaizit @jakobj

Issue #325, opened 2015-09-09
Drop most of the graphing utils


Debilski:
We should at least move much of the stuff in graph.py out of the pelita core (and maybe put it into some helpers library for the demo players and say that it is unsupported).

Algorithms which are supported (ie. if we want to provide a simple shortest path finder) need a better API and more documentation and tests.

otizonaizit:
We can keep the little we have.

PR #371, opened 2016-09-12
TST: Improved testing for A-star algorithm by adding cases from both sides.

DoriekeG:
Hi!

While writing the bots last week I noticed some weird behaviour, especially when they were moving from the upper right corner to the left side of the field using A*. The previous smoke test didn't cover this scenario, so I added some more rigorous tests to test the A* path searching from both sides of the field.

Dori

### DIFF
--- a/test/test_graph.py
+++ b/test/test_graph.py
@@ -187,14 +187,18 @@ class TestAdjacencyList(unittest.TestCase):
     def test_a_star(self):
         test_layout = (
         """ ##################
-            #0#.  .  # .     #
-            #2#####    #####1#
-            #     . #  .  .#3#
+            #02.# .  # .  #  #
+            #   ###    ####1 #
+            # ### . #  .  ##3#
+            #                #
             ################## """)
         universe = CTFUniverse.create(test_layout, 4)
         al = AdjacencyList(universe.free_positions())
-        # just a simple smoke test
-        self.assertEqual(14, len(al.a_star((1, 1), (3, 1))))
+        #Test distance to middle from both sides
+        self.assertEqual(11, len(al.a_star((1, 1), (7, 2))))
+        self.assertEqual(12, len(al.a_star((2, 1), (7, 2))))
+        self.assertEqual(14, len(al.a_star((16, 1), (7, 2))))
+        self.assertEqual(15, len(al.a_star((15, 1), (7, 2))))    
 
     def test_path_to_same_position(self):
         test_layout = (


Debilski:
Thank you, Dori. Great find. I wonder how many bots suddenly improve with your fix. :)

Debilski:
Wow, that thing is completely broken.

    from pelita.graph import AdjacencyList
    from pelita.datamodel import CTFUniverse


    def show_shortest_path(layout):
        """ Finds a shortest path between bot 0 and 1 and places
        food pellets on each field. """
        uni = CTFUniverse.create(layout, 2)
        al = AdjacencyList(uni.free_positions())
        path = al.a_star(uni.bots[0].current_pos, uni.bots[1].current_pos)
        uni.food = path
        return uni.compact_str

    layout1 = """
    ##################
    #  1  #    #  0  #
    #     #    #     #
    # #####    ##### #
    #                #
    ##################
    """

    layout2 = """
    ##################
    #  0  #    #  1  #
    #     #    #     #
    # #####    ##### #
    #                #
    ##################
    """


    print(show_shortest_path(layout1))

    ##################
    #..1  #    #  0..#
    #.    #    #    .#
    #.#####    #####.#
    #................#
    ##################

    print(show_shortest_path(layout2))

    ##################
    #  0..#....#  1  #
    #.....#.  .#  ...#
    #.#####.  .#####.#
    #.......  .......#
    ##################

NelleV:
\o/

rougier:
Just for the records: https://github.com/HarrietInc/elegant-scipy-submissions/issues/20
But I've no idea if this is more efficient (time) than A*.

otizonaizit:
Well, I tried to run a git bisect to see if this is a regression or the a_star has always been broken. Unfortunately the bisect does not work out of the box because of some rebasing that has happened in the meanwhile. But looking at the history of `pelita/graph.py` it seems clear that the implementation hasn't changed much in the last 5 years. @DoriekeG's PR #372 seems to fix it, even the snippet @Debilski posted here runs correctly once you merge #372 . 
I propose that we merge #372 and #371 and add @Debilski's script as a test (where we test that the path lengths are the same in both cases, right? It doesn't need to be the identical path, what matters is only length). 
Of course it would be nice to also add some more generic tests ;)

Debilski:
+100 for merging, after we have added additional tests with the script. Same goes for bfs probably. If profiling tells us that the difference is not too bad, we might even drop A\* and just use bfs with a big warning that one should rather use an external library for that anyway.

otizonaizit:
So, #378 merges #371 and #372 and add the additional test we were discussing here. Adding tests to bsf can be done on a different PR. I don't agree with referencing any external libraries. We should have a decently performing shortest path alg in Pelita. Does not need to be rocket science. If current a_start and bsf can be improved, or one of them dropped, we should just do it.

Debilski:
For the record: We *did* have the impression that it might be wrong earlier: https://github.com/ASPP/pelita/issues/325

DoriekeG:
Since A* is technically a form of bfs, I do not see the need for another implementation. 

Something that might improve the A* implementation would be to use nodes instead of lists. So each square that is examined has a variable parent that points to its parent node. That way the path can be reconstructed faster.

Debilski:
On large mazes, our bfs implementation performs 2x slower than a quick Python rewrite of the A* on the wikipedia page. I‘ll do a few more tests, though and then we‘ll decide. I‘d think when we can optimise bfs then it should indeed be enough for the purpose of getting started with a bot.

otizonaizit:
>Something that might improve the A* implementation would be to use 
>nodes instead of lists. So each square that is examined has a variable 
>parent that points to its parent node. That way the path can be 
>reconstructed faster.

That is a very good idea. Do you want to give it a try? We have two 
implementations in pelita/graph.py. Ideally we would have only one, 
which is fast and reliable. @DoriekeG: do you have time for a PR on 
this? :)

DoriekeG:
I will give it a go!

otizonaizit:
>I will give it a go!
Great! Put tests and new implementation in the same PR, so that Travis 
can do his job ;-)

Debilski:
Soeren Laue sent me a layout where it still fails with the improved A*:

    layout3 = """
    ########
    #1#... #
    #.#.#0 #
    #...   #
    ########
    """

    print(show_shortest_path(layout3))
    ########
    #1#... #
    #.#.#0 #
    #...   #
    ########

DoriekeG:
Could this be due to the backtracking algorithm? If it goes through all the still open nodes and opens (5,1) in seen before it opens (5,2) and finds the target, it has (5,2) very recently in the seen list. It will then instead of backtracking through (5,3) backtrack through (5,1) and return the wrong path. If this is indeed the problem, it should also be solved by using nodes.

PR #372, opened 2016-09-12
FIX: Changed A-star algorithm so that the history is taken into account

DoriekeG:
Hi!

I found a little bug in the A* algorithm, where it didn't take the path already travelled into account. This caused it to have some weird behaviour when starting in the upper right corner with walls in the way (see pull request for new test). 

Dori

### DIFF
--- a/pelita/graph.py
+++ b/pelita/graph.py
@@ -276,7 +276,7 @@ class AdjacencyList(dict):
             else:
                 seen.append(current)
                 for pos in self[current]:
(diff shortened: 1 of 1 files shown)


Debilski:
Something about that algorithm still feels wrong to me (now the heapq values overestimate the distance) which I believe should not be done with A\*.

I cannot find a counterexample, though. Would you say the rest is correct? Otherwise I’ll just switch to the algorithm from Wikipedia https://en.wikipedia.org/wiki/A*_search_algorithm#Pseudocode

PR #384, opened 2016-09-14
BF: Allow tournament to find pelitagame from any directory.

Debilski:
Small bug fix. Should be done more thoroughly in libpelita, though.


Debilski:
So …, who is going to review it then?

PR #386, opened 2016-09-14
Fix CI engine and move call_pelitagame to libpelita

Debilski:
Slight problem: call_pelitagame is POSIX only and Python 3.5+ so it might not be the best fit for libpelita which is supposed to be more generic. (CI and tournament can be POSIX and 3.5 only.)

Also: recursion: pelitagame imports libpelita and then finds a function that executes pelitagame … should we move it somewhere else?

Future work:
- The cli could be better defined and exposed as function args
- Add a flag that prints directly to stdout instead of capturing it
- The zmq socket could be passed as an environment variable to avoid ambiguity in the popen call

### DIFF
--- a/contrib/ci_engine.py
+++ b/contrib/ci_engine.py
@@ -129,17 +129,80 @@ class CI_Engine:
             the indices of the players
 
         """
+        import zmq
+
+        ctx = zmq.Context()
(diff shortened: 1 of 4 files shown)


Debilski:
Do not merge. Found a bug: call_pelitagame can not run concurrently because it will always use the same zmq socket.

Debilski:
Replaced by #428.

PR #388, opened 2016-09-18
setup.py

Debilski:
It is unclear to me how we should deal with the calls to Popen, ie. how do we know which script to call on all systems.


Debilski:
Doing a `python setup.py install` seems to work. (Afterwards `pelita team1 team2` works but it is still unclear if it is as flexible as before.)

@otizonaizit You want to give some initial feedback?

otizonaizit:
>@otizonaizit You want to give some initial feedback?

will do, thanks!

pelita/libpelita.py
```diff
@@ -274,7 +275,7 @@ def run_external_viewer(subscribe_sock, controller, geometry, delay):
     if delay:
         viewer_args += ["--delay", str(delay)]
 
-    tkviewer = os.path.join(os.path.dirname(sys.argv[0]), "tkviewer.py")
+    tkviewer = os.path.join(os.path.dirname(sys.argv[0]), "pelita-tkviewer")
```
otizonaizit: hmm, this is dangerous (in Debian for example `*.py` and `*.pyc` files get installed in different directory structures, and the `dirname` call may return the wrong dir)… can't this be addressed like in line 80 with a script in the `pelita.scripts` namespace?

Debilski: Yes. Seems I missed that one.

otizonaizit:
How do you start a game after installing pelita? Does this still allow usge without installation, or should we change the docs to reflect this and tell people to use `pip install -e .` or a variant thereof?

Debilski:
Without installation one can call the scripts module directly (`python -m pelita.scripts.pelita_main`). Of course, this may produce false positives if pelita is also installed and one is not in the main directory of the repo, so I am not sure we should really advertise it except for developers.

otizonaizit:
Yes, I agree. So, can you add some installation instructions in the 
docs? And what would be our suggestion for upgrading the pelita 
installation (in case we have to make some hot-fix during the school)? 
Just uninstall and install again or some cooler way?

Debilski:
Not really sure. I mean we could add a flag that runs `pip install --upgrade git+git://github.com/ASPP/pelita` and takes care of everything, but I think it would not be right.

Do we still want to require them to download the repo though? It might really get a little tricky if they want to debug things inside Pelita itself.

Debilski:
What about our versioning: Should we add something like versioneer or should we keep doing it manually?

otizonaizit:
we can keep on doing it manually, I guess.

Debilski:
TODO:

- [ ] Fix the versioning (make git version also PEP compliant)
- [ ] Think about making a namespace package for the sane pelita players
- [ ] Rename `tournament` -> `pelita_tournament` so that it does not clutter the site-packages
- [ ] *or* put everything back in `pelita` but make `pelita` a namespace package (everything currently in there moves to subpackage to `pelita.libpelita`)

Debilski:
@otizonaizit I still need a bit of ideas of how to do it correctly with the players and the tournament.

Should they go back to the pelita namespace for easier installation? But I am still reluctant to do so because they only have a one-directional relationship to the main programs (and in the case of the tournament it is also that this is Python 3.5+ only).

Or can they stay outside but be renamed `pelita_players` and `pelita_tournament`?

Debilski:
Ok, I think I have it in a state that is workable (needs a little more fine-tuning in setup.py maybe).

Debilski:
We are on PyPI (with a dev version) https://pypi.org/project/pelita/ 🎉

Debilski:
@otizonaizit I think I’m done here. Happy to clear up any uncertainties and merge.

otizonaizit:
I'll review on Monday. Hold on until then ;-)

otizonaizit:
can you update the documentation in this PR too? The docs still refer to git cloning and starting ./pelitagame …

Debilski:
git grep pelitagame -- doc/

Nothing.

otizonaizit:
ouch, I was checking the online rendered version, which, of course, 
can't contain the changes in this PR ;-)

Debilski:
Nevertheless, we should upgrade the docs again once we know what we want to have as the canonical setup.

> Am 19.08.2017 um 12:40 schrieb Tiziano Zito <notifications@github.com>:
> 
> ouch, I was checking the online rendered version, which, of course, 
> can't contain the changes in this PR ;-)
>

Debilski:
> Am 23.08.2017 um 11:42 schrieb Tiziano Zito <notifications@github.com>:
> 
> @otizonaizit requested changes on this pull request.
> 
> Very good job! It works well and it is much cleaner than before! Thanks!
> 
> I have a minor comment, that should be addressed before merging:
> 
> Now that we have namepsaces, the duality pelita.player and pelita.players is confusing, and indeed in the docs there is at least a wrong import:
> 
> doc/source/groupN/test/test_my_player.py:from pelita.player import RandomPlayer
> this one I noticed, but there may be others hidden somewhere… Couldn't we rename pelita.players to something else? Or. maybe, just bite the apple and merge pelita.player and pelita.players even if they logically are two different things? I think it would make the code organization easier to follow and less confusing. And, a search/replace players ⟶ player seems easier than the other way round ;-) I vaguely remember that we had this discussion again, but I notice only now that this is indeed an issue.
> 
> 

Yeah, I also find it confusing. I think we can somehow merge it. Most things in player.py are based upon AbstractPlayer, so that should be okay. On the other hand, what I don’t feel too good about is the `from pelita.player import SimpleTeam` which always sounds a bit wrong-worded to me. We could put that somewhere else, too (and change the name from SimpleTeam to something better).

But for now, I guess, I can make `pelita.player` a proper module and move everything there (unless you have a better name for `pelita.players`.)


> Another minor issue: I would just make the PyYAML dependency explicit like pyzmq. Installing pelita should always install the whole thing, so there is no point to have the "extra" dependency. Right now installing with pip install . does not install PyYAML by default. As a consequence the tournament does not work, but running py.test gives errors that don't show the real problem.
> 

The tournament needs numpy as well, though (not strictly; it only is used to help in printing the KO round). And the tournament is posix and Python 3.5+ only. So this may only superficially help. (We should fix the py.test then and maybe for the next release get rid of the numpy dependency.)

> Would it make sense to install a py.test config file that automatically ignores doc, so that one can run py.test with no arguments to run all tests?
> 
> 
I’ll have to think about it later.

otizonaizit:
+1 for merging `players` and `player`
-1 for moving and renaming `SimpleTeam`: we can do that, but in another PR and later. I have some ideas about it, but I don't think it's worth it to do it now.

I would then say to just add `numpy` and `PyYAML` to the dependencies and declare that we just support Python 3.5+. Why not? Getting rid of `numpy` is a valuable goal, but not so fundamental that it needs to happen very soon. The tournament is an integral part of the library, so if that part has more strict compatibility requirements than the rest, it has to drive the requirements of the whole library. I don't see a future where you want to split it out and have different Python version requirements or dependencies for the client, the server and/or the tournament…

Debilski:
Ok. I trust you in having the big picture there. Then let’s make the tournament 1st class. :)

To be honest, we _could_ downgrade the code to be 3.4 compliant (just new list syntax). But I am very ok with raising the base line here for future innovations.

Debilski:
`pelita/player.py` has now moved to `pelita/player/base.py` and all from `pelita/players/*` is now in `pelita/player/`.

`base.py` is an implementation detail until we find a better structure and is currently not used anywhere. Every class is imported into `__init__.py`, additionally the `SANE_PLAYERS` list is defined there as well.

Debilski:
> Would it make sense to install a py.test config file that automatically ignores doc, so that one can run py.test with no arguments to run all tests?

Actually, it is a good thing when it also runs `doc/source/groupN/test/test_my_player.py`. At least there is no harm in doing so.

Debilski:
All requested changes are added …

PR #389, opened 2016-09-25
A star with Nodes

DoriekeG:
Hi all,

I changed A star so it now uses Nodes to store the path. The Node class is in the utils folder. I haven't had time yet to check if it speeds up the algorithm a lot and if it fixes the bug Rike mentioned in #371 

Please let me know if you have any comments. Thanks! 

Dori

### DIFF
--- a/pelita/graph.py
+++ b/pelita/graph.py
@@ -3,6 +3,7 @@
 import heapq
 from collections import deque
 from .utils.Node import Node
+import time
 
 class NoPathException(Exception):
     pass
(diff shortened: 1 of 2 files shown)


Debilski:
Cool. Many thanks. It seems to be a tiny bit slower but that is not a problem.
I also have another test that we can attach to this PR. Do you want to rebase this branch upon the upstream master and push again? (And also possibly allow maintainers to append commits: https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/ )

Debilski:
And possibly rename the file `Node.py` to a lowercase version…

Debilski:
:( still fails:

    lay = """
    ################################################################
    #0#                #    #         #                     #   #  #
    # ######### ######           #              #           ###    #
    # #            #   ######## ## ## #  #      #           #   #  #
    #   ############   # # #  #  #         ## ###############      #
    # #            # ### # #     # ###  ##        #         #### ###
    # ####### #### #   #   #  #  #       #                         #
    # #   1      #     ###   ##### ##      ############# ###########
    # #   #      # #   #   #     # ##    #   #                     #
    #    ######################### ##    ## ######### ##############"""

    print(show_shortest_path(lay))
    ################################################################
    #0#                #    #         #                     #   #  #
    #.######### ######           #              #           ###    #
    #.#            #   ######## ## ## #  #      #           #   #  #
    #...############   # # #  #  #         ## ###############      #
    # #.......     # ### # #     # ###  ##        #         #### ###
    # #######.#### #   #   #  #  #       #                         #
    # #   1...   #     ###   ##### ##      ############# ###########
    # #   #      # #   #   #     # ##    #   #                     #
    #    ######################### ##    ## ######### ##############

Maybe we should really just replace it with the wikipedia version.

Debilski:
For reference, here is a bfs search that traverses a layout and checks all the path lengths:

    import heapq
    def check_all_paths(layout, start):
        uni = CTFUniverse.create(layout, 4)
        al = AdjacencyList(uni.free_positions())
        visited = set()
        to_visit = []
        heapq.heappush(to_visit, (0, start))
        while to_visit:
            dist, current_node = heapq.heappop(to_visit)
            if current_node in visited:
                continue
        
            path = al.a_star(start, current_node)
            uni.bots[0].current_pos = start
            uni.bots[1].current_pos = current_node
            uni.food = path
            print(uni.compact_str)
            print(start, current_node)
            assert dist == len(al.a_star(start, current_node))
            assert dist == len(al.a_star(current_node, start))
        
            visited.add(current_node)
            neighbors = al[current_node]
            for n in neighbors:
                if n not in visited:
                    heapq.heappush(to_visit, (dist + 1, n))

otizonaizit:
For example of implementations, maybe we should have a look at: http://www.redblobgames.com/pathfinding/a-star/implementation.html ?

I think the shortest path algorithm needs to be decently fast, because we don't want people to be hampered by it: they'll have enough other things that slow down calculations to take care of while developing a bot. @DoriekeG, @Debilski: what do you think? If @DoriekeG's version with nodes still does not work correctly, maybe we should start from scratch? In the link above there are several example implementations of shortest path algorithms... maybe worth to profile them and take the fastest?

Debilski:
https://github.com/ASPP/pelita/pull/390

Debilski:
I believe, I have found a fix: The problem is that the new estimates are calculated from the old estimates (i.e. Manhattan dist) and not from the real value. So we need to keep track of the proper distances as well.


        for pos in self[current]:
            new_dist = dist_so_far[current] + 1
            if pos not in dist_so_far or new_dist < dist_so_far[pos]:
                new_estimate = new_dist + manhattan_dist(target, pos)
                dist_so_far[pos] = new_dist
                heapq.heappush(to_visit, Node(new_estimate, pos, current_node))
 
I’ll now let the different implementations battle in order to find the speediest one.

Debilski:
@otizonaizit I suppose we should go with the last one: https://gist.github.com/Debilski/f66d10c5f16934d45c1ee37fe14371a0

Debilski:
Closed in #394.

PR #390, opened 2016-09-26
ENH: Use the wikipedia version of the A* algorithm.

Debilski:
Needs more tests and docs.


Debilski:
It’s also faster than what we had…

Debilski:
Closing. Is not the fastest implementation.

Debilski:
Closed in #394.

PR #393, opened 2016-09-26
RF: A* implementation from redblobgames.

Debilski:
Seems to be faster than the wikipedia implementation.


Debilski:
Closed in #394.

PR #394, opened 2016-09-27
Feature/astar fix

Debilski:
Fixed (?) with the help of http://www.redblobgames.com/pathfinding/a-star/implementation.html#python


Debilski:
Closes #393, #389, #390.

DoriekeG:
Looks good, it seems like this is indeed better than trying to patch up the existing code.

PR #404, opened 2017-08-04
Proof of concept for network players.

Debilski:
(On top of the setup.py changes, sorry.)

A call to

    remote# pelita_player --remote players/SmartRandomPlayer.py tcp://0.0.0.0:12345

binds the given zmq socket and waits for incoming match requests. For each request it opens a subshell which connects to the local game master and plays the game.

Local players can then play against the networked player with

    local# pelita my_player remote:tcp://server:12345

Current problems:

  * local does not know its IP address, therefore remote does not know how to connect to the local game master. This may either be given by the local user (straight forward to implement but won’t work with firewalls then) or the whole setup has to be restructured to use zmq addresses (this means we will have to reuse the sockets, I think).

Do not merge yet but it is better than nothing.

### DIFF
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -96,19 +96,6 @@ def default_players():
     from ..player import SANE_PLAYERS
     return sorted(SANE_PLAYERS, key=lambda m: m.__name__)
 
-def start_logging(filename):
-    if filename:
-        hdlr = logging.FileHandler(filename, mode='w')
-    else:
-        hdlr = logging.StreamHandler()
-    logger = logging.getLogger('pelita')
-    FORMAT = \
-    '[%(relativeCreated)06d %(name)s:%(levelname).1s][%(funcName)s] %(message)s'
-    formatter = logging.Formatter(FORMAT)
-    hdlr.setFormatter(formatter)
-    logger.addHandler(hdlr)
-    logger.setLevel(logging.DEBUG)
-
 def geometry_string(s):
     """Get a X-style geometry definition and return a tuple.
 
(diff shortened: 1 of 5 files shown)


Debilski:
When given a remote: spec, SimpleServer will now connect to the given
address as a zmq.DEALER (non-remote will still bind as a zmq.PAIR).

On the remote player side, the --remote flag opens a corresponding
zmq.ROUTER which waits for incoming requests.

  * If the request comes from an unknown DEALER, the program checks
    whether the message is a request and creates a new local zmq.PAIR and
    a subprocess of itself running the actual player code (which
    receives the address of the local zmq.PAIR.
    The zmq id of the incoming DEALER is kept in a database together
    with the zmq.PAIR socket.

  * Messages from known DEALERs and zmq.PAIR sockets are simply
    forwarded to their corresponding partner.

Debilski:
Closes #368.

Debilski:
Please note that it is now also trivial to run players in a container:

```
# Dockerfile
FROM ubuntu

RUN apt-get update
RUN apt-get install -y git python3-pip

RUN pip3 install -U pip
RUN pip3 install --user git+git://github.com/Debilski/pelita@feature/network-connections

EXPOSE 31144

CMD /root/.local/bin/pelita-player --remote RandomPlayer tcp://0.0.0.0:31144
```

    sudo docker build -t pelita_player_1 .
    sudo docker run -p 12345:31144 -d pelita_player_1
    sudo docker run -p 12346:31144 -d pelita_player_1

    pelita remote:tcp://127.0.0.1:12345 remote:tcp://127.0.0.1:12346

Debilski:
Needs a bit of green-threadification (broken and slow players break the system) but we can add this afterwards.

otizonaizit:
Let us review this once #388 is merged…

PR #407, opened 2017-08-09
ENH: Let the controller watch whether Tk is still alive.

Debilski:
This allows us to fail quickly when no $DISPLAY is set (otherwise it will hang). Before merging, I will need to think about the consequences though (maybe there is a situation where we don’t want to kill the game even though Tk has exited).

### DIFF
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -390,6 +390,7 @@ def main():
                 publisher = channels["publisher"]
                 game_config["publisher"] = publisher
                 viewer = libpelita.run_external_viewer(publisher.socket_addr, controller.socket_addr, geometry=geometry, delay=delay)
+                controller.watch_procs.append(viewer)
                 libpelita.run_game(team_specs=team_specs, game_config=game_config, viewers=viewers, controller=controller)
             else:
                 libpelita.run_game(team_specs=team_specs, game_config=game_config, viewers=viewers)
--- a/pelita/simplesetup.py
+++ b/pelita/simplesetup.py
@@ -408,7 +408,7 @@ class ExitLoop(Exception):
 class SimpleController:
     """ Sets up a simple Controller to interact with GameMaster. """
 
-    def __init__(self, game_master, address, reply_to=None):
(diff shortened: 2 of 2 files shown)


otizonaizit:
Can you rebase this on top of #388 or, better, rebase it on top of master once #388 is merged? Right now this PR contains a lot of unreleted commits and it is difficult to isolate the relevant changes…

Debilski:
I think the relevant change is https://github.com/ASPP/pelita/pull/407/commits/ef31cca1ab9ea778711197870c08fddb1bcf650a but I will rebase of course later.

Debilski:
If you find one such case, then we can keep it. But even right now, if one presses quit in the UI, this tells the controller to finish as well.

Having to wait 5 secs when the UI dies and then having to press CTRL-C is too long (and I experience this regularly because the DISPLAY is often not set in tmux).

otizonaizit:
But why modify the controller? Can't we catch the error in the viewer 
and exit cleanly? If DISPLAY is not set it fails with:

_tkinter.TclError: no display name and no $DISPLAY environment variable

we could catch this and exit like if one had pressed QUIT, no? 
[Reporting the error, of course]

Debilski:
> Am 23.08.2017 um 13:43 schrieb Tiziano Zito <notifications@github.com>:
> 
> But why modify the controller? Can't we catch the error in the viewer 
> and exit cleanly? If DISPLAY is not set it fails with:
> 
> _tkinter.TclError: no display name and no $DISPLAY environment variable
> 
> we could catch this and exit like if one had pressed QUIT, no? 
> [Reporting the error, of course]


There may be other errors, of course, so I am not sure we should really hard-code the Display variable error. And I think it makes sense in the controller because it should monitor its subprocesses anyway otherwise you have a bunch of still-running daemons of players or viewers in the background, which is a common problem with the new network players as well (it is easily fixable though by simply monitoring the subprocesses that one popened).

I am on a ferry right now. :) I can check later if this fix is a must (because it also fixes other bugs) or a mere annoyance (then it can wait).


>

PR #428, opened 2017-09-19
Make a complete game of pelita executable from libpelita

Debilski:
Replaces #386.

This moves the run_match logic from the tournament to libpelita. Currently this spawns a new subprocess with the given command line arguments and captures the final state as well as stdout, stderr. This might make it thread-safe, although there are no tests for this yet.

The disadvantage is the slightly awkward work-arounds that go with running an external process and having basically no feed-back while the function is running.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,6 @@ language: python
 sudo: false
 matrix:
   include:
-    - python: "3.3"
-      env: PYZMQ="pyzmq"
-    - python: "3.4"
-      env: PYZMQ="pyzmq"
     - python: "3.5"
       env: PYZMQ="pyzmq"
     - python: "3.6"
--- a/pelita/libpelita.py
+++ b/pelita/libpelita.py
@@ -2,11 +2,14 @@
 
 from collections import namedtuple
 import contextlib
+import io
+import json
 import logging
 import os
 import shlex
 import subprocess
 import sys
+import uuid
 
 import zmq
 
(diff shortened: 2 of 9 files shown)


Debilski:
Still big TODO: `call_pelita` opens a subprocess of `pelita_main` which in turn opens subprocesses of `pelita_player` for the respective players. If something in `call_pelita` crashes, we want to automatically terminate all subprocesses. Unfortunately, terminating `pelita_main` circumvents the `atexit` handlers (I think) and therefore `pelita_main` cannot inform its own children and `pelita_player` will wait forever in the `SimpleClient` loop for some zmq input.

Debilski:
@otizonaizit If you don’t see anything glaringly wrong with my preexec_fn=os.setsid (and the Windows equivalent), I would like to merge this (after running all tests on Windows myself again).

otizonaizit:
I am not sure I fully understand the logic, but why `setsid` instead of `setpgid`? (But maybe the question is due to my misunderstanding of the intent)…

Debilski:
Basically because `setsid` was the first thing that worked. I am not sure if I totally understand the difference between the two calls. Would I run `setpgid` also just once on the first child and have all its children inherit it automatically? Or would it need to be set on all the grand-children separately?

otizonaizit:
I'd guess that setpgid should be called on the grandparent. Then, all children should automatically be created wit the same process-group id and you can kill all of them by killing the PGID. I guess, but only trying will show if that works…

Debilski:
Ok, but I don’t want to kill the grandparent (which would be the Python code that directly runs `call_pelita`, eg. the tournament or some CI script). There I just want to catch an exception and then recursively terminate whatever `call_pelita` ran in a subprocess.

Debilski:
Merging still not possible: My smart way of switching between zmq and stdout/stderr io does not work on Windows (`poll.register(proc.stdout.fileno(), zmq.POLLIN)`)…

Issue #49, opened 2011-09-13
on_left_side property


pberkes:
property that returns True when the agent starts on left side of maze:
useful for example in:

```
    forbidden = west if self.on_left_side else east
    if self.maze_nodes[pos] in self.border_nodes and action == forbidden:
        return stop
```

sappelhoff:
I have some trouble adding a test.

When adding  the following to `test/test_datamodel.py`:

    def test_on_west_side(self):
        bot = Bot(1, (6, 6), 1, (3, 6), current_pos = (1, 1))
        assert bot.on_west_side

I get this error upon running `pytest test/test_datamodel.py`:

`E       AttributeError: 'Bot' object has no attribute 'on_west_side'`

Furthermore, when I go to `datamodel.py` and change e.g., the `is_harvester` method to `is_harvester2`, the tests still pass despite the fact that the `is_harvester` method should be missing ...

Any idea why this happens @Debilski?

Debilski:
@sappelhoff Answering in #429

Debilski:
Closed in #430

PR #429, opened 2017-10-23
added "on_left_side" property for bots

sappelhoff:
fix issue #49

### DIFF
--- a/pelita/datamodel.py
+++ b/pelita/datamodel.py
@@ -115,6 +115,16 @@ class Bot:
         return self.in_own_zone
 
     @property
+    def on_west_side(self):
+        # Find team_border of maze
+        if 0 in self.homezone:
+            border = max(self.homezone) + 1
+        else:
+            border = min(self.homezone)
+        # We are on_west_side, if current_pos is left of team_border
+        return self.current_pos[0] < border
+
+    @property
     def is_harvester(self):
         return not self.is_destroyer
 
--- a/pelita/graph.py
+++ b/pelita/graph.py
@@ -287,7 +287,7 @@ class AdjacencyList(dict):
                     heapq.heappush(to_visit, (priority, next))
         else:
             # no target found
-            raise NoPathException("BFS: No path from %r to %r." % (initial, target))
+            raise NoPathException("a_star: No path from %r to %r." % (initial, target))
 
         # Now back-track using seen to determine how we got here.
         # Initialise the path with current node, i.e. position of food.
--- a/test/test_datamodel.py
+++ b/test/test_datamodel.py
@@ -67,6 +67,12 @@ class TestBot:
         assert bot.is_harvester
         assert not bot.in_own_zone
 
+    def test_on_west_side(self):
+        bot_west = Bot(1, (1, 1), 0, (3, 6), current_pos = (1, 1))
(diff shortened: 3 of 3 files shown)


Debilski:
My personal take: This should at most be a property on `AbstractPlayer` but not in the datamodel API. But I have always been against this suggestion by @pberkes anyway (I think no-one should ever need it) so I am not the person to decide this one. :)

pberkes:
Hi @sappelhoff, thanks for the contribution! I needed this while writing my own bots.

Rike, can you clarify why you don't want this method, or where it should go? If there is a better way, can you please add a comment to issue #49 and close it?

pelita/datamodel.py
```diff
@@ -115,6 +115,13 @@ def is_destroyer(self):
         return self.in_own_zone
 
     @property
+    def on_left_side(self):
+        # True if current_pos is left of team_border
+        return (self.current_pos[0] < ((sorted(self.homezone)[1] + 1)
```
Debilski: `sorted(self.homezone)[1]` is equivalent to `max(self.homezone)`, right?

Debilski:
@pberkes I find it too specific (and it kind of relies on how we set up the initial rules). But yeah, the rules are not very likely to change.

@sappelhoff can you rename it to `on_west_side` or similar. And please add a test.

pelita/datamodel.py
Debilski: And that nested if is a little convoluted. I think a verbose multi-line if is better here.

sappelhoff: @Debilski, I thought I would have to import math or numpy for a max function ... just realized now that it's not necessary to import anything for that :-) 

Next commit will be more verbose

Debilski:
> I have some trouble adding a test.
> 
> When adding the following to test/test_datamodel.py:
> 
> def test_on_west_side(self):
>     bot = Bot(1, (6, 6), 1, (3, 6), current_pos = (1, 1))
>     assert bot.on_west_side
> I get this error upon running pytest test/test_datamodel.py:
> 
> E AttributeError: 'Bot' object has no attribute 'on_west_side'
> 
> Furthermore, when I go to datamodel.py and change e.g., the is_harvester method to is_harvester2, the tests still pass despite the fact that the is_harvester method should be missing ...
> 
> Any idea why this happens @Debilski?

Hi, @sappelhoff, I think this is because you still run the tests against the old pelita. You have to install the dev version from the local git repo with the `-e` flag:

    cd pelita
    pip install -e .

sappelhoff:
> cd pelita
pip install -e .


Thanks Rike, that was it 👍 

I made the changes and opened a new pull request with a clean commit history and proper branch name. See pull request #430. Please just close this one, I kind of messed up rebasing/squashing the commits and don't know what happened.

pberkes:
@Debilski did you want @sappelhoff to move the code to `AbstractPlayer`?

Debilski:
@pberkes On second thought it only makes sense to have it on `AbstractPlayer` if you are only going to use it with your current players and never on the other bots (which may be exactly the use case but I am not sure).

I am open to postpone this decision for a more general API restructuring.

Debilski:
Closed in favour of #430

PR #430, opened 2017-10-24
NF/UT: added on_west_side property for bots and a test for it

sappelhoff:
fixes issue #49 

see pull request #429

### DIFF
--- a/pelita/datamodel.py
+++ b/pelita/datamodel.py
@@ -111,6 +111,16 @@ class Bot:
         return self.homezone[0] <= self.current_pos[0] <= self.homezone[1]
 
     @property
+    def on_west_side(self):
+        # Find team_border of maze
+        if 0 in self.homezone:
+            border = max(self.homezone) + 1
+        else:
+            border = min(self.homezone)
+            # We are on_west_side, if current_pos is left of team_border
+            return self.current_pos[0] < border
+
+    @property
     def is_destroyer(self):
         return self.in_own_zone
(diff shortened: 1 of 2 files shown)


pelita/datamodel.py
```diff
@@ -111,6 +111,16 @@ def in_own_zone(self):
         return self.homezone[0] <= self.current_pos[0] <= self.homezone[1]
 
     @property
+    def on_west_side(self):
+        # Find team_border of maze
+        if 0 in self.homezone:
+            border = max(self.homezone) + 1
```
Debilski: The `return` for this case is missing.

Debilski:
Hey, @sappelhoff before you do any changes to this PR please give me a bit of time to think it through. Maybe it should indeed not be on `Bot` because it takes into account the universe (and `Bot` is just a dumb object). So it is better on either `CTFUniverse` or `AbstractPlayer` but I am still figuring out which one.

Debilski:
@sappelhoff Let’s keep it as it is but please add tests for all possible cases (if you still want to work on it). We can think about putting it somewhere else when we do some future refactoring.

sappelhoff:
Hi @Debilski, I fixed the missing return case that you spotted earlier, added the property to the bot-class docstring, and added a comprehensive test for the property.

I think adding it for now and later refactoring it is a good solution.

Debilski:
Thanks a lot.

PR #431, opened 2017-10-26
Give warning if creating layout without food (fixes #418)

### LINKED ISSUE #418: Give a warning, when a layout without food is defined?

ageorgou:
Fixes #418. I know that we were burnt by this, so I thought I would address it...

This just checks that a layout includes food and, if not, gives a warning. I've put the warning in the factory method that creates the layout rather than in the GameMaster - I don't know what you'd prefer. Since there are several tests with no food, I've also added a pytest.ini to suppress the warnings from them and keep the output cleaner (I couldn't see any other customisation for py.test already in the repository).

Edit: Just saw the format for marking commits, sorry!
Edit2: Failing on 3.3 seems strange, although I see that configuration uses a different pytest version (2.9.2 vs 3.x in later ones). I don't see anything about how warnings are handled in the changelog, though.

### DIFF
--- a/pelita/game_master.py
+++ b/pelita/game_master.py
@@ -4,6 +4,7 @@ import abc
 import random
 import sys
 import time
+from warnings import warn
 
 from . import datamodel
 from .datamodel import Bot, CTFUniverse
@@ -17,6 +18,10 @@ class PlayerTimeout(Exception):
     pass
 
 class PlayerDisconnected(Exception):
+    pass
+
+class NoFoodWarning(Warning):
+    """ Warning about a layout with no food. """
     pass
 
 class GameMaster:
@@ -157,6 +162,12 @@ class GameMaster:
             #: sight distance of the noise
             "noise_sight_distance": self.noiser and self.noiser.sight_distance
         }
+
+        # Check that both teams have food, and raise a warning otherwise
+        for (team_id, food_count) in enumerate(self.game_state["food_to_eat"]):
+            if food_count == 0:
+                warn("Layout contains no food for team {}.".format(team_id),
+                     NoFoodWarning)
 
     @property
     def game_time(self):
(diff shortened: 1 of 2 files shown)


Debilski:
That’s very cool. I’ve never used the warnings with pytest but I assume that they will only be suppressed for the Pelita code itself and not for the user made tests, right?

Don’t worry about Python 3.3, the mental baseline at the moment is 3.5.

Debilski:
Two remarks:
  * I think we should already warn when one of the teams has no food (because this will also finish the game).
  * Personally, I think there may be valid use cases for a *universe* without food but for a *game* it does not make much sense so it should rather be in `GameMaster` (maybe we could then also get rid of the ignore file).

We can wait a bit if someone else has a strong opinion on this, though.

ageorgou:
So, I think pytest -at least in its current version- records all warnings (except for deprecation warnings) and reports them after the tests have run, whether they were caused by the test itself or the code it calls. I think there is a decorator to suppress warnings from a particular test, but then you'd have to wrap all the tests, unless I missed something.

The setting in the file makes it so that pytest doesn't report this particular warning, just to avoid a mass of messages at the end from tests having no food. The warnings are still raised (and captured if required, like in the test) though. There is a command line option for doing the same thing, but the file seemed more robust.

Agreed about moving the check to GameMaster - didn't know if it was better to "catch" it early or be more specific. If done there, it's also easier to check for either team not having food, too.

Debilski:
@ageorgou Given that no-one other opinions arrived: If you are able to add the changes and update this PR, please go ahead.

ageorgou:
@Debilski Sure, I will hopefully get to it later today.

ageorgou:
Rebased to clean up changes and mark commits. Check is now in GameMaster; I've added it at the end of the initialisation, since it is a warning, not an error, and the rest of code would run anyway. I'm not entirely sure that the team number displayed in the warning message is guaranteed to be consistent with other team indices!

Debilski:
Merged. Thanks a lot.

PR #435, opened 2017-11-23
Initial work on new, simpler API for pelita

jni:


otizonaizit:
I am merging this to the branch simplify-player. This should make it easier to work collaboratevely. We can just make PR against that branch and merge without fear of breaking stuff.

Debilski:
See follow-up in #435 .

PR #436, opened 2017-11-24
Rename AdjacencyList to Graph

Debilski:
We can also decide whether we want to get rid of the `dict` inheritance there.


jni:
My opinion is that we want to replace our ad-hoc graph with a NetworkX graph...

otizonaizit:
Hmm, I am just a bit worried of having networkx as a dependency, and networkx is nt the only graph-handling library out there. Our Graph is pretty simple and easy to use, we don't want our student to go and read alsnoso networkx documentation… Maybe our Graph class should be able to export to networkx if asked to do so, but, I think this can be done already, so it is just a matter of documenting it, is this correct @Debilski ?

Debilski:
Yes, exporting to networkx just needs to be documented a little more. I’ll append a commit for this.

Issue #391, opened 2016-09-26
Rename ‘def factory’


Debilski:
This is confusing for people who have never heard of a factory method. Calling it `def team` makes some sense (this is also the name that many groups like to choose for their own factories) but collides with the name `team` for the default group module.

One of the `team`s has to go:

```
./groupN/team:team
```

Ideas?

PR #437, opened 2018-01-15
UX improvement: The default factory is now called ‘team’

Debilski:
Players are now required to export a method

    def team():
        return SimpleTeam(…)

from their modules.

NB: This is a breaking change!

### DIFF
--- a/README.tournament.md
+++ b/README.tournament.md
@@ -11,7 +11,7 @@ All configuration is done with a yaml file:
     speak: false
     seed: null
     bonusmatch: True
-    
+
     teams:
       - spec: RandomPlayer
         members:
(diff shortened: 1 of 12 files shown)


Debilski:
I am open for other suggestions, also with respect to the group repo layout, which is currently

    group0/team/__init__.py

The folder `team` could therefore also be renamed. This is just a convention for the tournament and configured in yaml anyway.

Debilski:
E.g.

    def players()

See. #391

Debilski:
Or should it be a singular name to stress the AI code behind?

Debilski:
API change in a minor version. Where do I have to donate for absolution?

PR #460, opened 2018-03-06
Proof of concept for a minimal PyQt5 UI

Debilski:
Features:

- zmq polling in a separate thread (in the single-threaded Tk we have to query every few milliseconds whether new data has arrived)
- proper anti-aliasing on Linux and macOS (not tested on Windows yet)
- the possibility for a useful png export (currently behind the `--export` cmd arg)
- cubic bezier curves for the ghosts
- pink food
- longer startup times
- no buttons (I did not bother to create them until now and also Qt buttons are slightly in the uncanny valley on macOS)

It’s just a proof of concept and I don’t expect this to be merged or be used as the standard client. Tk is more on the batteries-included side of things and probably more tested. Also, the license of PyQt5 is GPL so we might want to keep it separate. However, we might use it as a base for exporting pngs in a proper and convenient way.

image


pelita/libpelita.py
```diff
@@ -399,7 +399,9 @@ def channel_setup(publish_to=None, reply_to=None):
     yield { "publisher": publisher, "controller": controller }
 
 
-def run_external_viewer(subscribe_sock, controller, geometry, delay):
+def run_external_viewer(subscribe_sock, controller, geometry, delay, viewer=None, export=None):
+    if viewer == None:
```
jni: `if viewer is None` is the Pythonic way. ;)

Debilski: I think I usually write `is None` but patching this method was such a bad hack already that I mentally blacked out that thing I guess. :)

jni:
So cool! The pink food is a critical feature. =D

I'm impressed with the brevity of the code. I think if you can paint a proper maze (this could be done by painting a brick pattern, or just making black squares of enough size), it's worth merging.

Debilski:
Yeah, painting the maze properly is not much of a issue but I will have to look into implementing some caching for that. Currently it does a live redraw which will probably be slower if we have some connected walls (simple bricks don’t look good – there must be some common outline).

Qt comes with built-in scaling and transformations so we don’t have to write it. Apart from that it is not too different from the Tk version actually. Just more minimal. :)

Debilski:
Hmm, I think we should have some shading for the wall background (or more fluid lines). Otherwise it just looks too similar to the Berkeley one.
image

image
(http://ai.berkeley.edu/contest.html)

pelita/ui/qt/qt_viewer.py
```diff
@@ -0,0 +1,316 @@
...
+class QtViewer(QMainWindow):
+    def __init__(self, address, controller_address=None,
+                       geometry=None, delay=None, export=None,
+                       *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if export:
+            png_export_path = Path(export)
+            if not png_export_path.is_dir():
+                raise RuntimeError("Not a directory: {png_export_path}")
```
jni: shouldn't this get formatted?

Debilski: It’s a commit named ‘WIP’. This will be rebased eventually. Sorry for the noise. ;)

PR #466, opened 2018-04-01
Pelita for Excel

Debilski:
Scientists spend more and more time writing, maintaining and debugging Excel files. While techniques for doing this efficiently have evolved, only few scientists have been trained to use Excel. As a result, instead of doing their research and publishing Word docs, they spend far too much time fighting auto format and removing erroneously-pasted cells …

To show those scientists that they are not alone, we now have Pelita for Excel:

[Pelita for Excel](https://vimeo.com/262719894 "Pelita for Excel")

https://vimeo.com/262719894


jni:
That video is incredible!

keszybz:
I'm speechless.

PR #467, opened 2018-04-01
ENH: New approach for the simplified player API.

Debilski:
The general idea is that we give up on the object oriented interface and instead have the user supply a move function (or perhaps several for the individual bots) that receives a universe/state and a storage dict.

The universe/state object should contain all necessary information to evaluate a state. The representation will be kept as simple as possible (a simple print should show all available information to help with observability) and we will export all the previous helper methods on AbstractPlayer as functions.

Additionally, we initialise an empty storage dict which we pass to the move function. The storage dict can be used to share data between the individual players and rounds and is persistent for a whole game.

### Implementation strategy

The goal will be to get to an immutable universe/game_state representation for the user side that can be queried with external accessor functions. Users can write their own accessor functions and they will look exactly the same as our functions. (This is not possible with the OO interface, or at least not for non-OO people.)

The goal is not to change the datamodel API. This can follow eventually but it is not necessary just yet.

Our starting point for the representation will be created from the `_to_json_dict` transformation of `CTFUniverse` (which is the same format that we use for sending through zmq) and will consist of dicts, lists and simple types only.
Starting from this representation forces us to not look back and re-use the datamodel API.
Inside the `Team` class, we will have a function that performs a few ‘cosmetic adjustments’ (removing redundancy, improving the format). If deemed good, we may want to backport those changes to the zmq wire-format as well but again, this will be a second step.

Open questions and TODOs:

* Right now, the storage dict is stored in the `Team` class. We should decide whether we want to allow reusing a `Team` for more than one game and what happens with the storage dict in this case.

* In the OO version, it feels more natural to have people use `self.rng` for the seeded `random` module (and it will automatically be persistent between rounds). How should we solve this with a functional API?

* Do we want to have reserved keys in the storage dict? Those could be used for returning additional information (the move function itself only returns a direction). Most prominently, this could be used for having the bots talk (perhaps also for rng state?) but it could also be useful for some kind of UI overlay for debugging (we could have the players print additional information on the Tk interface in development mode).

* How do we pass the bot’s own index to the move function? In a separate argument or inside the datadict?

* Will we keep the `set_initial` phase?

### What is already done?

I added a new `Team` class to `pelita.player` which replaces the old `SimpleTeam` (the one that worked with `AbstractPlayer`s). This class can be used in parallel with players using `SimpleTeam`. A small (working but not yet useful) example of how it could look like eventually has been given in `demo/demo_team.py`.

The API for the player is to be added to `pelita/player/player_functions.py` (for now). There we should now implement all those methods from `AbstractPlayer` and datamodel that we still feel are useable. We might want to start with a more minimalistic set of methods though and not provide an accessor for everything.

### DIFF
--- a/demo/demo_team.py
+++ b/demo/demo_team.py
@@ -1,0 +1,15 @@
+
+import random
+
+from pelita.player import Team
+from pelita.player.player_functions import legal_moves
+
+def move1(datadict, storage):
+    legal = legal_moves(datadict)
+    return random.choice(legal)
+
+def move2(datadict, storage):
+    return (0, 0)
+
+def team():
+    return Team("My Team", move1, move2)
--- a/pelita/player/__init__.py
+++ b/pelita/player/__init__.py
@@ -5,6 +5,8 @@ from .base import (StoppingPlayer, SteppingPlayer, SpeakingPlayer,
                    RoundBasedPlayer, MoveExceptionPlayer, InitialExceptionPlayer,
                    DebuggablePlayer)
 
+from .team import Team
+
 from .RandomPlayers import RandomPlayer, NQRandomPlayer
 from .FoodEatingPlayer import FoodEatingPlayer
 from .SmartEatingPlayer import SmartEatingPlayer
--- a/pelita/player/player_functions.py
+++ b/pelita/player/player_functions.py
@@ -1,0 +1,6 @@
+
+# Player API
+
+def legal_moves(datadict):
+    # TODO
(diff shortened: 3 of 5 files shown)


Debilski:
@jni If you want to work on this during your flight as a starting point (@otizonaizit this is what we discussed in last week’s meeting – or at least what I thought we did), I’ll merge to ASPP:api-rework.

jni:
@Debilski great! I've checked it out. =)

PR #472, opened 2018-06-09
ENH: First draft for error-handling capabilities in the zmq protocol

Debilski:
pelita_player will now intercept when load_team fails and return an
error message object to the zmq socket. The ZMQConnection has been
adapted to close itself and raise a connection error on retrieval.

As a quick fix, the error message is returned from set_initial as the
team name so that it can be shown in the UI.

Further updates should make this a bit more flexible and refined.

Positive changes: A typo on the command line or a syntax error in a player will not freeze the UI for 3 seconds anymore.

Needs some more testing before merging. Also I might still rename some of the exceptions.

### DIFF
--- a/pelita/scripts/pelita_player.py
+++ b/pelita/scripts/pelita_player.py
@@ -35,13 +35,26 @@ def with_sys_path(dirname):
 
 
 def make_client(team_spec, address):
-    team = load_team(team_spec)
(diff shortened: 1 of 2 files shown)


Debilski:
Merging now. Too useful not to have it.

Issue #476, opened 2018-06-23
Reformulate one sentence in the rules


fmaussion:
In last year's presentation, the rules state:

"Winning: A game ends when either one team eats all of the opponents’ food pellets, or the team with more points after 300 rounds."

This can be misunderstood as "eating all the food makes you win". It could be reformulated as:

"Winning: A game ends when either one team eats all of the opponents’ food pellets, or after 300 rounds. The team with more points wins."

I think it would be nice to have the rules as a section in the documentation also. I can take care of this later this week.

PR #484, opened 2018-08-17
DOC/ENH: add a "rules" section to the docs

fmaussion:
Fixes https://github.com/ASPP/pelita/issues/476 

Related: did you guys ever tried to change the winning condition to eating all the enemy's food first regardless of how many times you die? I guess this would change the strategies quite a bit, but for good or for bad?

### DIFF
--- a/doc/source/rules.rst
+++ b/doc/source/rules.rst
@@ -14,8 +14,9 @@ returns to its starting position (as a harvester). *5 points*
 are awarded for eating an opponent.
 
 **Observations**: Bots can only observe an opponent’s exact position, if they
-or their teammate are within 5 squares of the opponent bot. If they are
-further away, the opponent’s positions are noised.
+or their teammate are within 5 squares of the opponent bot (in Manhattan
+distance, see :ref:`maze_distances`). If they are further away, the opponent’s
+positions are noised (see :ref:`noisy_positions`).
 
 **Timeout**: each Player only has 3 seconds to return a valid move. If it
 doesn’t, a random move is executed and all later return values are
--- a/doc/source/rules.rst
+++ b/doc/source/rules.rst
@@ -1,0 +1,26 @@
+==========
+Game rules
+==========
+
+Each team owns two Bots, and each Bot is controlled by a Player. Bots in
+their homezone are called "Destroyers", and "Harvesters" when in the
+enemy’s homezone.
+
+**Eating**: when a Bot eats a food pellet, the food is permanently removed and
+*one point* is scored for that Bot’s team.
+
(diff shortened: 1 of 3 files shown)


Debilski:
We probably thought about it (I think sometimes when giving the presentation, I still forget to explain that eating all food doesn’t automatically win the game for you). It might make some of the sitting-on-food strategies invalid, so this could be an advantage.
On the other hand, it is unclear why we should be having points then in the first place.

fmaussion:
> forget to explain that eating all food doesn’t automatically win the game for you

This is the purpose of this PR ;)

otizonaizit:
I think winning by eating all the food would restrict the number of strategies, in the sense that then it becomes "just" a matter of finding the sequence of food pellets which can be reached through the minimal number of steps, i.e. it turns it into a mathematical problem (which may already have a solution?). Giving points for the killing (and a lot at that) makes the solution space way more complicated, and makes it also more fun to watch, because students will then try to kill ;)

doc/source/rules.rst
```diff
@@ -0,0 +1,26 @@
...
+**Eating**: when a Bot eats a food pellet, the food is permanently removed and
+*one point* is scored for that Bot’s team.
+
+**Eating another Bot**: when a Bot is eaten by an opposing destroyer, it
+returns to its starting position (as a harvester). *5 points*
+are awarded for eating an opponent.
+
+**Observations**: Bots can only observe an opponent’s exact position, if they
+or their teammate are within 5 squares of the opponent bot. If they are
+further away, the opponent’s positions are noised.
```
otizonaizit: We should say here that 5 squares does *not* mean that they are reachable in 5 moves, i.e. it is manhattan distance and not distance on the graph (students are often confused by this). And we should specify what kind of noise is added (I think it is a uniformly distributed integer between -5 and +5 on each coordinate? @Debilski ?)

doc/source/rules.rst
```diff
@@ -0,0 +1,26 @@
...
+
+**Timeout**: each Player only has 3 seconds to return a valid move. If it
+doesn’t, a random move is executed and all later return values are
+discarded. 5 timeouts and you’re out!
+
+**Winning**: the game ends when *either* one team eats all of the opponents’
+food pellets, *or* after 300 rounds. The team with more points at the end of
+the game wins.
```
otizonaizit: This section could be slipt in: **End**, the game ends... and **Winning**. This way it is also visually clear that winning the game or causing a game over is not the same....

fmaussion: Good idea. Done.

PR #489, opened 2018-08-20
ENH: Report illegal moves on the command line.

Debilski:
Previously, it was only reported as a timeout. Now we also specify the reason when it came from an illegal move.

### DIFF
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -43,6 +43,11 @@ class ResultPrinter(pelita.viewer.AbstractViewer):
                                   game_state["timeout_teams"][universe.bots[bot_id].team_index],
                                   universe.bots[bot_id].team_index,
                                   bot_id))
+            elif reason == "illegal_move":
+                sys.stderr.write("Illegal move -> Timeout #%r for team %r (bot index %r).\n" % (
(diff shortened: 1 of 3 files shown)


pelita/scripts/pelita_main.py
```diff
@@ -43,6 +43,11 @@ def print_bad_bot_status(self, universe, game_state):
                                   game_state["timeout_teams"][universe.bots[bot_id].team_index],
                                   universe.bots[bot_id].team_index,
                                   bot_id))
+            elif reason == "illegal_move":
+                sys.stderr.write("Illegal move -> Timeout #%r for team %r (bot index %r).\n" % (
```
otizonaizit: Can you add this to the list of rules in the documentation? Something in the form of "A timeout is received if the bot takes too long to respond or if it requests an illegal move". Otherwise it is unclear what the word timeout means in this context. Of course it would be nice to report both the illegal move request as well as the random move that gets assigned instead, but, OK, that is for another time...

Debilski: I thought about adding the move but reporting both will clutter the internal API too much since I am only flagging it with the `reason` key. This really is better done at a later point.

Debilski: Added commit with clarification.

otizonaizit:
For me you can merge as soon as Travis is done

PR #493, opened 2018-08-22
ENH: Added --stop-after flag which stops the game in the given round.

Debilski:
Currently the first step of the round will be executed. This should be fixed in a future revision.


Debilski:
Needs to be tested on other OS before merge.

PR #502, opened 2018-08-24
initialize _track and _eaten internal variables in Bot object

otizonaizit:
... so that we can use them in tests


otizonaizit:
Not sure, I think it would break the current implementation in Team. It seems to work this way.

otizonaizit:
the Travis errors are unrelated and the corresponding fix is coming in pelita_template in a minute

PR #503, opened 2018-08-24
Make Graph a subclass of Userdict

otizonaizit:
Given that moving Graph to be its own class (right thing to do) requires
some work at the moment, we make it a subclass of UserDict, so that at
least we can properly implement `__copy__` This is needed because
the default `dict.copy()` behaviour is shortcutting `__copy__`.
Anyway, this mess needs to be cleaned up and Graph should be its own
thing.

### DIFF
--- a/pelita/graph.py
+++ b/pelita/graph.py
@@ -1,7 +1,7 @@
 """ Basic graph module """
 
 import heapq
-from collections import deque
+from collections import deque, UserDict
 
 
 class NoPathException(Exception):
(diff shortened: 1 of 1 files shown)


otizonaizit:
merging anyway, because Travis failure is unrelated (@Debilski : can you fix this?)

Debilski:
The travis failure is pelita_template, no?

otizonaizit:
no

PR #506, opened 2018-08-24
add stop, i.e. (0,0), to the list of legal_moves

### LINKED ISSUE #505: Can we add (0, 0) a.k.a stop to Bot.legal_moves?
I am not sure why this is excluded, but it seems an oversight. Stopping is always a legal move, so why shouldn't it be on the list? This will make some special casing in the code not needed anymore...

otizonaizit:
Fixes #505

### DIFF
--- a/pelita/player/team.py
+++ b/pelita/player/team.py
@@ -262,11 +262,10 @@ class Bot:
 
     @property
     def legal_moves(self):
-        """ The legal moves that the bot can make from its current position.
-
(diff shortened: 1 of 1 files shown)


otizonaizit:
Note that we have two failing test. One is unrelated (@Debilski : can you fix that?) the other one is in pelita_template and it's expected, given that that test assumes that stop is not a legal move.

PR #507, opened 2018-08-24
Remove other and team from Bot

### LINKED ISSUE #504: `Bot.team` and `Bot.other` are redundant?
I am not sure if we discussed about it, but in the new API the Bot object holds a reference to
the other bot (`Bot.other`) and to the team (`Bot.team`). Given that both these things are actually already avaiable as `game.team` and `game.team[1-turn]` we can get rid of them? If they are needed internally then they should be prefixed with a `_`. All other public attributes and methods of `Bot` make sense and can be presented to the students. Removing them is not going against the principle of the flat API we decided for, so I think they can go.

otizonaizit:
Fixes #504

### DIFF
--- a/pelita/player/team.py
+++ b/pelita/player/team.py
@@ -275,14 +275,6 @@ class Bot:
         return legal_moves
 
     @property
-    def other(self):
-        """ The other bot in the team.
-        """
-        other_index = (self.bot_index + 2) % 4
-        return self._bots[other_index]
-
-    @property
-    def team(self):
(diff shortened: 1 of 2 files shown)


fmaussion:
Thanks for the info!

I will be away from my keyboard for the next three weeks, so I won't be able to contribute unfortunately. I plan to use Pelita again the upcoming semester, so I'm curious about what you guys will report from the summer school ;)

PR #509, opened 2018-08-24
BF: Catch ZeroDivisionError in fps calculation.

Debilski:
Python’s time is either inexact on Windows or Pelita is just super fast there …


otizonaizit:
@Debilski : could you fix the Travis error? It is becoming hard to review pull requests: they all look like they are failing!

Debilski:
I fixed it yesterday already …

PR #511, opened 2018-08-25
Allow unicode in team names.

### LINKED ISSUE #409: Allow Greek script in team names.

Debilski:
**This is the future that liberals want!!**

unicodepelita

Closes #409

### DIFF
--- a/pelita/scripts/pelita_player.py
+++ b/pelita/scripts/pelita_player.py
@@ -13,6 +13,7 @@ import signal
 import string
 import subprocess
 import sys
+import unicodedata
 
 import zmq
 
@@ -60,24 +61,23 @@ def make_client(team_spec, address, color=None):
 
 
 def check_team_name(name):
-    # Team name must be ascii
-    try:
-        name.encode('ascii')
-    except UnicodeDecodeError:
-        raise ValueError('Invalid team name (non ascii): "%s".'%name)
     # Team name must be shorter than 25 characters
-    if len(name) > 25:
+    name = name.strip()
+    for n in name:
+        print(unicodedata.category(n))
+    # for the length check, we’ll remove a few characters
+    # otherwise we’ll penalise teams with accents
+    unicode_len = len(u''.join(ch for ch in name
+                                    # don’t count combining characters
+                                  if unicodedata.combining(ch) == 0
+                                    # or control characters
+                                  and not unicodedata.category(ch).startswith('C')
+                                    # or separatos
+                                  and not unicodedata.category(ch).startswith('Z')))
+    if unicode_len > 25:
         raise ValueError('Invalid team name (longer than 25): "%s".'%name)
-    if len(name) == 0:
+    if unicode_len == 0:
         raise ValueError('Invalid team name (too short).')
-    # Check every character and make sure it is either
-    # a letter or a number. Nothing else is allowed.
-    for char in name:
-        if (not char.isalnum()) and (char != ' '):
-            raise ValueError('Invalid team name (only alphanumeric '
-                             'chars or blanks): "%s"'%name)
-    if name.isspace():
-        raise ValueError('Invalid team name (no letters): "%s"'%name)
 
 
 def load_team(spec):
--- a/test/test_pelita_player.py
+++ b/test/test_pelita_player.py
@@ -76,7 +76,7 @@ class TestLoadFactory:
 
 class TestLoadTeam:
     def test_simple_module_import_forbidden_names(self):
-        names = ["", " ", "-", "∂", "0" * 26]
+        names = ["", " ", "0" * 26]
         for idx, name in enumerate(names):
             modules_before = list(sys.modules.keys())
             with tempfile.TemporaryDirectory() as d:
@@ -91,7 +91,7 @@ class TestLoadTeam:
                     load_team(spec)
 
     def test_simple_module_import_allowed_names(self):
-        names = ["a", "a a", "0" * 25]
+        names = ["a", "a a", "0" * 25, "-", "∂"]
         for idx, name in enumerate(names):
             modules_before = list(sys.modules.keys())
             with tempfile.TemporaryDirectory() as d:


otizonaizit:
I don' think it is a good idea. What if the characters are not available to TK? This pull requests invites trolls to come up with stupid ideas, so I think it's better to leave it a ascii-only names

Debilski:
Hadn’t you requested this yourself? But since we accept any unicode in say(), I don’t see much of a difference here.

jni:
fwiw I'm in favour of Unicode characters. ASCII is a little bit exclusive. You don't have to explicitly announce that it works.

Debilski:
We can always restrict the allowed Unicode categories some more, if there turn out to be obvious displaying issues.

jni:
> would you say that Python has to accept module names with the hyphen character because is exclusive not to do so? For example you can have a module with your full surname ;)

I'd be more interested in getting `from scikit-image import segmentation` to work. =D

More seriously, Nick Coghlan [mentioned](https://ncoghlan-devs-python-notes.readthedocs.io/en/latest/python3/questions_and_answers.html) inclusivity as a specific reason for universal unicode support in Python 3:

> we did not see any other way to ensure Python remained a viable development platform as developer communities grow in locations where English is not the primary spoken language.

(Not that I agree with everything in that essay but I liked this bit.)

> remember that the decision to restrict to ascii **was** taken because a student troll was misusing Unicode and finding it funny to see how the game broke

I was not there when this happened, and I agree that this is a good reason to be careful with Unicode support. Having said that, If `.say` supports it, it's interesting that the team names should be an exception.

Anyway, I do agree that this is a low priority, but did not see a good reason to block the PR given that the work is essentially already done.

otizonaizit:
>>remember that the decision to restrict to ascii wastaken because a student troll was misusing Unicode and finding it funny to see how the game broke
>
>I was not there when this happened, and I agree that this is a good reason to be careful with Unicode support. Having said that, If `.say` supports it, it's interesting that the team names should be an exception.
>
>Anyway, I do agree that this is a low priority, but did not see a good reason to block the PR given that the work is essentially already done.

Well, `.say` supports it because @Debilski needed to print greek letters from his bots in Nikiti. I would argue that we have to remove unicode support from `.say` too. Just imagine what happens if someone writes the character that switches the display to right-to-left, or the invisible-separator character, or the compose-character or one of the other 1000 crazy unicode things. We can support unicode when someone has the time to learn how to support unicode without breaking everything. I don't think it's worth it and the potential risk and distraction is high.

otizonaizit:
By the way, the screenshot @Debilski posted in this issue is exactly the reason I want to keep on forbidding this and even extend ascii-only strings to the `say` method, see #513

jni:
> Just imagine what happens if someone writes the character that switches the display to right-to-left, or the invisible-separator character, or the compose-character or one of the other 1000 crazy unicode things.

I would have assumed that all of these things affect only the string being printed and not every subsequent string printed afterwards? Surely the default is to "sandbox" the team names / sayings? Or it should be simple to turn on? If I can change the UI text by typing something in a box, that seems to be a problem not with sanitising input, but with the security model altogether?

otizonaizit:
>I would have assumed that all of these things affect only the string being printed and not every subsequent string printed afterwards? Surely the default is to "sandbox" the team names / sayings? Or it should be simple to turn on? If I can change the UI text by typing something in a box, that seems to be a problem not with sanitising input, but with the security model altogether?

Sure it is a problem with the security model. I don't think we know what would happen. There is not really any sandboxing going on. Pelita is not a banking app, and it's not even a video game for the masses. So no, there's almost no security.

The only security is that client and server run on different processes, but it would be possible (and relatively easy) for a client to spoof another client auth or to break the game in many possible ways. The server does a certain amount of input sanitation, but surely not enough.

Again, we are not hardening pelita, I am just trying to avoid student trolls. And we have had them, and they typically belong to the arrogant-assholes category, as it requires a certain dose of nit-picking to try and see if you can type random unicode things and have fun seeing pelita break down.

If there is any pelita-development time available, we have a long list of useful **and** urgent things to do. Proper Unicode support and hardening the security model are quite low in the priority list at the moment ;)

Debilski:
At least on OS X , the RTL markers were handled correctly by Tk though.

If we want to harden by character restrictions then we should go one step further and only allow ascii uppercase (to make it more 80s arcade like).

Debilski:
Found an example where it breaks:

    $ tail -2 pelita/player/FoodEatingPlayer.py
    def team():
        return SimpleTeam("The 🇮🇹 Food Eating Players", FoodEatingPlayer(), FoodEatingPlayer()

    $ pelita
    Replay this game with --seed 6320411448875762495
    Using layout 'layout_normal_without_dead_ends_089'
    Red team '1' -> 'Random Explorer Players'
    Blue team '0' -> 'The 🇮🇹 Food Eating Players'
    Exception in Tkinter callback
    Traceback (most recent call last):
      File "/Users/me/anaconda3/envs/pelita/lib/python3.7/tkinter/__init__.py", line 1702, in __call__
        return self.func(*args)
      File "/Users/me/anaconda3/envs/pelita/lib/python3.7/tkinter/__init__.py", line 746, in callit
        func(*args)
      File "/Users/me/Projects/pelita/pelita/ui/tk_viewer.py", line 161, in wrapped_fun
        fun(*args)
      File "/Users/me/Projects/pelita/pelita/ui/tk_viewer.py", line 144, in read_queue
        self.app.observe(data)
      File "/Users/me/Projects/pelita/pelita/ui/tk_canvas.py", line 697, in observe
        self.update(universe, game_state)
      File "/Users/me/Projects/pelita/pelita/ui/tk_canvas.py", line 367, in update
        self.draw_universe(universe, game_state)
      File "/Users/me/Projects/pelita/pelita/ui/tk_canvas.py", line 394, in draw_universe
        self.draw_title(universe, game_state)
      File "/Users/me/Projects/pelita/pelita/ui/tk_canvas.py", line 478, in draw_title
        rel_size = 1)
      File "/Users/me/Projects/pelita/pelita/ui/tk_canvas.py", line 24, in guess_size
        text_width = font.measure(display_string)
      File "/Users/me/anaconda3/envs/pelita/lib/python3.7/tkinter/font.py", line 154, in measure
        return self._tk.getint(self._call("font", "measure", self.name, *args))
    _tkinter.TclError: character U+1f1ee is above the range (U+0000-U+FFFF) allowed by Tcl

otizonaizit:
Can we kill this pull request?

otizonaizit:
No reactions to my call for kill. So killing!

PR #515, opened 2018-09-05
Layout passed to setup_test_game should always be a string

jakobj:
To avoid confusion there should be only one way to pass the layout to `setup_test_game`. It's easiest to pass this always as a string and add the `food`, `bots`, `enemy` parameter as arguments to `setup_test_game`.


Debilski:
I fear it might make `setup_test_game` a little too complicated with even more keywords (that react on each other in a particular way) but if everybody thinks it is okay then we can change.

PR #526, opened 2018-09-19
ENH: Do not print output for flite.

### LINKED ISSUE #521: flite stderr should be discarded in tournament mode
When `flite` complains on `stderr` during the tournament with  `speak` enabled, it gets printed on the console. This is a regression, because I know that the original call to flite would redirect both `stderr` and `stdout` to `/dev/null`. This does not happen anymore, and it is quite annoying.

Debilski:
Fixes #521 (crudely).


otizonaizit:
perfect, that is exactly what we need.

Issue #520, opened 2018-09-19
change new client API. From move(turn, game) to move(turn, game, state)


otizonaizit:
After the first run of the API on students it seems clear that having the state stored in the attribute `game.state` was a bad idea. We should move it to an additional argument to the `move` function, `None` by default. The `move` function should then return `next_move, state`. 

After this is done, it is just a matter of updating the documentation on `pelita_template` and update the demos. Not a lot of work.

Debilski:
While I agree that it is somehow the right thing to do, it feels a little unpythonic (in the sense that people are not really used to doing things that way) to me to require returning a tuple. Do we have any alternative ideas?

Also, for the record, the `Game` object will still have to be writeable, as it keeps the rng and the say variable. I assume this should stay that way, right?

otizonaizit:
> While I agree that it is somehow the right thing to do, it feels a little unpythonic (in the sense that people are not really used to doing things that way) to me to require returning a tuple. Do we have any alternative ideas?

No, short of mandating a global variable `STATE`, I don't see any better way. On top of that, I don't see the problem with returning a tuple. Why would it be unpythonic? What does that mean?
As usual, documentation makes all of the difference. If we use the new `state` argument consistently in the demos and document it properly in the template repo, it is going to create exactly zero problems.

> Also, for the record, the `Game` object will still have to be writeable, as it keeps the rng and the say variable. I assume this should stay that way, right?

Of course, there is no way to have a read-only object for real in Python anyway. But, from the point of view of the user, `Game` is going to be read-only now (`rng` is initiliazed before the user can interact with `Game`, and `say` is a method on `Bot`, so there's no reason for the user to assume that it is going to set some special attribute in `Game`). Again, I don't care much about the internals, I care about a consistent API for the end user.

Debilski:
> Am 19.09.2018 um 14:55 schrieb Tiziano Zito <notifications@github.com>:
> 
> While I agree that it is somehow the right thing to do, it feels a little unpythonic (in the sense that people are not really used to doing things that way) to me to require returning a tuple. Do we have any alternative ideas?
> 
> No, short of mandating a global variable STATE,
> 

Well, that would be even worse than what we have now.

> I don't see any better way. On top of that, I don't see the problem with returning a tuple. Why would it be unpythonic? What does that mean?
> 
My point is that people are not really used to have functions that return two arguments and it might lead to different problems, even if we document it well. I am all for trying it though, because it does make sense and I don’t see another good way either.

[...]

PR #527, opened 2018-09-19
ENH: Make state explicit in move

Debilski:
state is now an extra argument to the move function. In order to
properly update state, the move function now must return both the
move tuple and an updated state.:

    move(bot, state) -> (move, state)

Additionally, `move` receives a `Bot` instance of the current bot which is equivalent to the previous `game.team[turn]`.

Concept branch for #520 which requires additions (and discussions?) in https://github.com/ASPP/pelita_template before we should merge it. (Should we manually update the 2018 players so that they still run for tests or provide a compatibility mode?)

### DIFF
--- a/pelita/player/team.py
+++ b/pelita/player/team.py
@@ -58,7 +58,7 @@ class Team(AbstractTeam):
 
         #: Storage for the team state
         self._team_state = None
-        self._team_game = Game([None, None])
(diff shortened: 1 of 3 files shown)


otizonaizit:
Travis-CI are expected. Merging anyway.

Issue #519, opened 2018-09-19
Client-generated Exceptions are lost in the tournament


otizonaizit:
This is a serious flaw. It seems stderr is not captured anymore. The code is too messy and I couldn't find where this could be fixed. I'd prefer to print to stdout directly and store stderr on a separate file instead of using the logging system from Python...

otizonaizit:
Note that we want the Exceptions to be explorable directly **during** the tournament. So, if we choose to redirect `stderr` to a spearate file, that file should be flushed at the end of each game to be sure we can see what happened when a game is over because of a client failing. (It happened during the tournament and no one knows nor can figure out what happened).

otizonaizit:
Fixed in #528

PR #528, opened 2018-09-20
ENH: Store the tournament output next to the dumpfile.

Debilski:
Quick fix for #519

Stdout and stderr are captured in `call_pelita` but are just written to a variable that is then returned by `call_pelita`. So far, the tournament script would only present the stdout and stderr to the user, if `call_pelita` itself did not return a meaningful result. In the case of a player crashing, this would just trigger a normal disconnect and the other team would simply win, so it was never considered to be a situation where the tournament host would need to investigate.

This adds a fix to the tournament that will store stdout and stderr of `call_pelita` directly in a file, *after* `call_pelita` has run.

We could additionally add the feature that `call_pelita` writes directly to a file (or log), if requested, but unless we need live updates (`tail -f`), this fix should suffice.

### DIFF
--- a/pelita/libpelita.py
+++ b/pelita/libpelita.py
@@ -112,7 +112,7 @@ class BinRunner(ModuleRunner):
         return external_call
 
 @contextlib.contextmanager
-def _call_pelita_player(module_spec, address):
+def _call_pelita_player(module_spec, address, color='', dump=None):
     """ Context manager version of `call_pelita_player`.
 
     Runs `call_pelita_player` as long as the `with` statement is executed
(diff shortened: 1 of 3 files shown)


otizonaizit:
I assume that the `dump` name is unique? Just to be sure that it doesn't get overwritten.
Otherwise, I'd prefer for stderr and stdout to be flushed at every line, and not only at the end, but I did not understand the whole `PYTHONUNBUFFERED` thing in `call_pelita`, so I do not know how feasible it is.

otizonaizit:
>  save to a file in the code that `Popen`s the players themselves (that way, we could have different output files for the different teams).

Yes, that seems the best way to go!

Debilski:
> Am 20.09.2018 um 15:34 schrieb Tiziano Zito <notifications@github.com>:
> 
> save to a file in the code that Popens the players themselves (that way, we could have different output files for the different teams).
> 
> Yes, that seems the best way to go!
> 

It’s not super trivial though because the `Popen`ing of the players is not synchronised currently (and the way `call_pelita` does it only works because we gave up Windows compatibility for the tournament). But that part needs some cleaning up anyway.

otizonaizit:
looks good!

otizonaizit:
one travis-CI failure is still expected (`bot.track` usage in `pelita_template`)

PR #535, opened 2018-10-01
Add --long-help

Debilski:
Adds a `--long-help` flag to `pelita`.

The implementation is a slight hack because the feature does not exist in argparse. It uses a helper function `long_help` that will be `argparse.SUPPRESS` when `--long-help` is not set so that it will usually hide the advanced features from normal users.

Additionally, this commit does some cleaning up and gets rid of the `initial_delay` variable.


Debilski:
Any issues with having a --long-help for the advanced flags, @otizonaizit? That way they are at least documented.

PR #536, opened 2018-10-01
Controllable replay mode

### LINKED ISSUE #453: Make the replay/snapshots mode first class
Replay mode is broken since the last refactoring. It is possible to fix it badly with a few lines copy/pasta but I think we should do it properly and also add regular tests so that this oversight does not happen again.

Debilski:
Closes #453. (To be rebased after #535 has been merged.)

The ReplayPublisher now includes controller functionality, which means that the Tk viewer can now request the next frame, when it wants to.
This means that we can get rid of the more or less arbitrary sleep functions for proper synchronisation. Additionally, we can now pause the replay and step through it. (Stepping through rounds is not implemented yet.)

Some further restructuring of the init logic in `pelita_main` has also been done.

### DIFF
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -18,17 +18,40 @@ logging.root.manager.emittedNoHandlerWarning = 1
 _logger = logging.getLogger(__name__)
 
 class ReplayPublisher:
-    def __init__(self, publish_sock, replayfile):
+    def __init__(self, replayfile, publisher, controller):
         with open(replayfile) as f:
             self.old_game = f.read().split("\x04")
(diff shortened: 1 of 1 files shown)


otizonaizit:
can you do the rebase and check that everything is still fine? #535 has been merged

otizonaizit:
It would be easier to review without the noise from the `--long-help` thing...

Debilski:
Rebased.

Debilski:
It only works with the Tk viewer (well, any external viewer should work), so it is still some sort of a hack. But it gets rid of the `sleep` therefore it is better. :)

PR #542, opened 2018-11-12
First-class movie creation mode

Debilski:
_Please do not yet merge. The new code as well as the surrounding old Tk code needs some more polishing._

This adds the highly sought-after ‘automatic replay movie creation mode’ to Pelita, which uses Tk’s `Canvas.postscript` function to write a postscript representation. As the generated postscript depends on what is shown on the screen, this mode also activates a few changes in the UI representation.

1) The canvas size will be resized such that each square has a size of 24x24 (open for discussion),
2) the status part of the UI will be hidden during the replay as it cannot be transformed to Postscript (it is not a Canvas) and hiding it makes some of the calculations needed for proper resizing easier.

Usage:

    $ pelita --dump replay.json    # I think we should rename --dump though
    ...
    $ pelita --replay replay.json --snapshot-folder my_replay/
    $ ninja -C my_replay/
    $ vlc my_replay/movie.mp4

For personal convenience, I am creating a `ninja.build` automatically. Not strictly needed since we are only expected to convert it once (except for testing, where it is useful) and a bash script should do just fine. But parallelisation is nice. :)

Current drawbacks:

* The snapshot mode creates two ps files. For the header and the game canvas. Both of them need to be converted to a png with Image Magic with proper settings to have the right amount of anti-aliasing.
* It is not clear which font will be used, so it is potentially very ugly looking.
* The first few frames should not be snapshotted (at least not before the game knows the names of the teams).
* ffmpeg needs to know the output dimensions and introduces some further rescaling.
* If the game window is resized during snapshotting, things will get weird.
* Creating the movie has to be done with external tools.


Debilski:
* Converting on macOS is really slow…

Debilski:
Currently I would much prefer looking into a solution that just uses a Qt interface and directly saves the canvas to png. Converting ps to png takes maybe 10 seconds per frame which is far too much time.

Debilski:
Let’s go full svg. It looks very all right and is quite fast to export (ie. not noticeable). That means we can export it already while the tournament is running.

image

Debilski:
Additional open idea: Do we actually need mp4 video files? Maybe an animated Javascript canvas is the better solution for presenting the matches after the tournament anyway. Obviously low priority.

PR #543, opened 2019-01-22
DOC: Update copyright statement

Debilski:


keszybz:
My suggestion would be to simply drop all copyright lines. They serve no purpose.

Debilski:
Yeah, makes sense.

Debilski:
To be fixed with the doc update in core-2.0.

Issue #433, opened 2017-11-01
`GameMaster(blah)` should better be called `create_game`


Debilski:
create_game(layout=..., teams=..., rounds=...)

might be a clearer API for users than having to create a weird `GameMaster` object. OTOH, since we cannot hide the `GameMaster` this may lead to additional confusion…

jni:
But, mostly yes? I think this would be very valuable. Currently, debugging and profiling in pelita is hard, and these are core skills that we want them to develop during the tournament. Most in our group gave up trying, including the tutors!

Debilski:
Yes. There’ll be an easier way either way. :)

otizonaizit:
Yes, we are going tohave a pelita sprint and we'll address this kind of issues. Just, this specific issue doesn't make sense per se. By the way, since we have introduced a demo of a bot which uses the debugger, I think it is clear now how to use a debugger with pelita and it is not hard at all.
The profiling is a different beast. But I don't see how we can make it better, given the way profilers work in Python. The students of course can keep on use timeit and others manual ways of measuring time. It's not like profiling is impossible.
OTOH, I agree it would be nice to be able to use a proper profiler, but really I don't see how to do it...

Debilski:
> I think it is clear now how to use a debugger with pelita and it is not hard at all.

Depends where you’re coming from. If you’re used to writing code in an IDE (and many people in science around me are) then it is less obvious. (I don’t have a really good solution though atm.)

otizonaizit:
>>If you’re used to writing code in an IDE (and many people in science around me are) then it is less obvious.
>
>That's exactly right. It's also "impossible" to do e.g. `pudb.runcall()` and set a breakpoint from within the debugger.

But we explicitly refrain from teaching IDEs at ASPP! And we want them for once to see how the low-level debugger looks like. It's like teaching git on the command-line. Debugging pelita with the low-level debugger works already good enough. I really think we have more pressant issues to address, instead of serving the hypothetical IDE user. If people are good enough to use the debugger integrated into their IDE, they probably don't need to be our students in the first place.

The workflow we teach with the debugger is to set a breakpoint manually in the code, then enter the debugger and navigate the code with it. This workflow works today with pelita (documented since 2018, so it is already a big improvement). I seriously doubt anything more is needed for our purposes (I am not talking about scientific coding in general, I am just talking about teaching how to use a debugger at ASPP).

jni:
[...]

PR #555, opened 2019-02-23
Add same-process mode to run_game that allows for easier debugging and profiling

### LINKED ISSUE #545: How to profile?
Running `kernprof /path/to/pelita --null my_player.py my_player.py` fails because the players are run in subprocesses. Some guidance on how to profile players would be useful!

### LINKED ISSUE #552: Add function that runs a few games in the background similar to call_pelita
At the moment the syntax for `call_pelita` is a little too complicated to be useful in scripting as it only mirrors the command line interface (in fact what it does is to run `pelita` from the command line).
This is useful for the tournament because having the main game run in a subprocess helps us separate the environments but for students wanting to know how their code performs after having played `N` matches, it is too much information.

    for l in pelita.get_layouts():
        final_state = run_match(group1_move, group2_move, rounds=300, layout=l)

It should be possible to specify the move functions either as a function (in which case no pelita-player is spawned and it will all be in one process) or as a team_spec (eg. `group0/group0.py`).

Debilski:
This PR makes a few API changes to the `libpelita.run_game` function, eliminating the need to pass a `game_config` dict in favour of explicit arguments (with sensible defaults). This should make it easier to use `run_game` in bulk testing similar to the following snippet:

    for enemy in ["enemy_players/team1", "enemy_players/team2"]:
        layout = get_layout() # ...
        game_state = run_game([my_move, enemy], rounds=300, layout=layout)
        # ...

Additionally, the team arguments to `run_game` may now be either team spec strings (as one would use on the pelita command line and which run in a subprocess) as well as plain `move(bot, state)` functions which will run in the same process and therefore allow for debugging (there is no timeout by design) and profiling.

A simple script for profiling might look as follows (the FoodEatingPlayer could be any other enemy):

    import pelita
    import time

    @profile
    def move(bot, state):
        time.sleep(0.01)
        return (0, 0), state

    name, layout = pelita.layout.get_random_layout()
    pelita.libpelita.run_game([move, "pelita/player/FoodEatingPlayer.py"], rounds=30, layout=layout)

And could be run with `kernprof -v -l profile.py` to yield:

    Wrote profile results to profile.py.lprof
    Timer unit: 1e-06 s

    Total time: 0.684961 s
    File: profile.py
    Function: move at line 7

    Line #      Hits         Time  Per Hit   % Time  Line Contents
    ==============================================================
         7                                           @profile
         8                                           def move(bot, state):
         9        60     684431.0  11407.2     99.9      time.sleep(0.01)
        10        60        530.0      8.8      0.1      return (0, 0), state

Closes #545. Closes #552. Addresses issues mentioned in #433.

### DIFF
--- a/pelita/libpelita.py
+++ b/pelita/libpelita.py
@@ -347,19 +347,22 @@ def prepare_team(team_spec):
         address = "tcp://127.0.0.1"
     return TeamSpec(module, address)
 
-def run_game(team_specs, game_config, viewers=None, controller=None):
+def run_game(team_specs, *, rounds, layout, layout_name="", seed=None, dump=False,
+                            max_timeouts=5, timeout_length=3,
+                            viewers=None, controller=None, publisher=None):
+
     if viewers is None:
         viewers = []
 
     teams = [prepare_team(team_spec) for team_spec in team_specs]
 
-    server = SimpleServer(layout_string=game_config["layout_string"],
-                          rounds=game_config["rounds"],
+    server = SimpleServer(layout_string=layout,
+                          rounds=rounds,
                           bind_addrs=[team.address for team in teams],
-                          max_timeouts=game_config["max_timeouts"],
-                          timeout_length=game_config["timeout_length"],
-                          layout_name=game_config["layout_name"],
-                          seed=game_config["seed"])
+                          max_timeouts=max_timeouts,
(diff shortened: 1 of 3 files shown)


jni:
@Debilski so cool! Could you make a corresponding example in `pelita_template`? =)

PR #561, opened 2019-02-25
Remove StringIO context manager

otizonaizit:
it does not help to use the context manager for the StringIO object.

By removing it we save one level of indentation, which makes the code more
readable. The diff looks crazy because git has problems understanding that
the code has been simply dedented by one level of indentation, but indeed
no code has been altered apart from the first line of the diff where the
"with StringIO..." was.


Debilski:
image

PR #563, opened 2019-02-26
RF: Store universe in GameMaster.game_state and remove it from APIs

Debilski:
We store the dict representation of CTFUniverse in the game_state. Since the universe is only accessed in a few methods in GameMaster, we can live with a small number of conversions. The worst drawback, however, is that we use GameMaster.universe extensively in the tests, where conversions are more annoying but hopefully only temporary.

Removing the universe arguments in the AbstractViewer and AbstractTeam APIs should make integration with the new core rewrite easier. The API for SimpleTeam or AbstractPlayer have not been changed yet.

This patch does not necessarily have to go to master (the conversions might make it a little slower even) but it will make the compatibility interface to the new core-2.0 much nicer (which makes for easier testing). In any case, I would also apply it to core-2.0 later (or rebase).

### DIFF
--- a/pelita/game_master.py
+++ b/pelita/game_master.py
@@ -62,16 +62,17 @@ class GameMaster:
     def __init__(self, layout, teams, number_bots, game_time, noise=True, noiser=None,
                  max_timeouts=5, timeout_length=3, layout_name=None,
                  seed=None):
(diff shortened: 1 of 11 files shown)


Debilski:
Let’s not make things even more complicated. All new work is done in core-2.0.

PR #564, opened 2019-03-02
V2.0 Added tests and fixed some bugs in ´play_turn´

lschwetlick:


Debilski:
I’ll merge and then rebase onto our new branch `pelita-core-2.0`, which is less-conflicting branch name (also did some minor cleanups in that branch). I am going to write an email where I suggest everybody uses this branch now.

Debilski:
@lschwetlick What is your intention with the seedval? To me it seems quite superfluous.

lschwetlick:
I'm not testing everything on all layouts, so by seeding it I can reconstruct which layout stuff went wrong with. I guess I can also just print the layout name or something though...

Debilski:
I think that’s better. I don’t think your approach is actually working, is it? (You just print a random number?)

lschwetlick:
True... Now you say it... something mast have got lost there at some point. Sorry

Debilski:
No problem. As you suggest, I’d just print the layout name and if there turn out to be tests that need more elaborate handling of random numbers, we can look into fixing it in pytest somehow.

Debilski:
Added a solution for the randomness in #569.

PR #573, opened 2019-03-19
Add RemoteTeam class

Debilski:
Given the methods in `game.py`, a `RemoteTeam` behaves just as a `Team` class (it is generated with `make_team` but takes a string instead of a `move` method) such as `set_initial` and `get_move`). Underneath however, it starts a subprocess of `pelita-player` with the `team_spec` as an argument.
When the class is cleaned up, the `__del__` method is (hopefully) called, which will then terminate the subprocesses (in case this hasn’t happened yet).

NB: This means that test games that run a few rounds of `play_turn`, then store the `game_state` in a list for example and then start a new game for a few rounds etc. will have dozens or more of still-running `pelita-player` processes around, since the teams are saved inside `game_state` and will therefore not be garbage collected. We might want to add a `finalize_game_state` function at some point.

NB2: We probably want to register the running processes with `atexit` as well.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -1,17 +1,30 @@
 """This is the game module. Written in 2019 in Born by Carlos and Lisa."""
 
 import dataclasses
+import logging
 from random import Random
 import sys
 import typing
 
 from . import layout
 from .gamestate_filters import noiser
+from .player.team import make_team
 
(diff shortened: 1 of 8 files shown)


Debilski:
`pelita --null` games are also possible now!

Issue #487, opened 2018-08-20
Rules unclear: What happens when bot respawns and an enemy sits on the respawn position


Debilski:
Example with the Python 3.7 manual cmdline player (need a better name):

    $ cat manual.py
    from pelita.player import SimpleTeam
    from pelita.player.base import SteppingPlayer

    def __getattr__(name: str):
        players = [SteppingPlayer(steps) for steps in name.split(',')]
        return lambda: SimpleTeam("Stepping", *players)

And the layout:

    $ cat chainlayout
    ########
    # .   1#
    #2.  .3#
    #0   . #
    ########

The idea is to have 0 and 1 swap places and then 2 attacks 1. Now 1 will respawn but of course 0 sits there. There several options possible:

- a) 0 will be killed immediately as well and if there were a bot on 0’s respawn point then it would also be killed … (This would lead to infinite loops on layouts where the respawn point is not in the own homezone, so there needs to be the additional rule that a bot can only be killed once per turn.)

- b) 0 will be killed in the next *turn*, no matter whose it will be

- c) 0 will be killed at the beginning of its turn (Q: Will the turn be over then?)

- d) 0 will be killed when the attacker has its turn (Q: Before the move or will it be decided afterwards, i.e. is it only a kill when the attacker stays where it is?)

- e) 0 will not be killed but has a chance to escape in its next turn.

Example I: We play 8 rounds and the left team wins:

    pelita 'manual:^^>>>>>--,-------v-' 'manual:<<vv<<<---,---------' --rounds 8 --layoutfile chain_layout --fps 4

Example II: 9 rounds and a draw:

    pelita 'manual:^^>>>>>--,-------v-' 'manual:<<vv<<<---,---------' --rounds 9 --layoutfile chain_layout --fps 4

Example III: 0 bravely escapes:

    pelita 'manual:^^>>>>>-<,-------v-' 'manual:<<vv<<<---,---------' --rounds 9 --layoutfile chain_layout

Debilski:
Added a spec in 4232e70704bc24f35599bf4733e296efa2f08a49 . We may want to add some docs before closing this issue.

Debilski:
Closing, although the logic has changed again and docs are still missing. :)

PR #575, opened 2019-03-25
Improving the logic of the round and turn handling

Debilski:
The logic is as follows:

run_game is basically a loop with the following conventions:

    def run_game(**game_kwargs):

        game_state = setup_game(**game_kwargs)
        assert game_state['round'] == None
        assert game_state['turn'] == None

        while not game_state['gameover']:
            game_state = play_turn(game_state)

            assert game_state['round'] >= 0
            # round is always smaller than max_rounds
            assert game_state['round'] < game_state['max_rounds']
            assert 0 <= game_state['turn'] < 4

The round and turn in a `game_state` _in this loop_ will always signify the bot that has last moved. If no bot has moved yet, these are naturally `None`. The same state will be given to the viewers (this happens inside `play_turn`) to avoid confusion. A function `check_final_move` is added to properly check whether another move can be played or whether we are in the last turn of the final round.

A bot will however see the _current_ round and turn. I.e. turn == the bot’s own index.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -115,6 +115,9 @@ class GameState:
 
 def run_game(team_specs, *, max_rounds, layout_dict, layout_name="", seed=None, dump=False,
              max_team_errors=5, timeout_length=3, viewers=None):
+    """ Run a match for `max_rounds` rounds.
+
+    """
 
     if viewers is None:
         viewers = []
@@ -262,6 +265,18 @@ def prepare_viewer_state(game_state):
 
 
 def play_turn(game_state):
+    """ Plays the next turn of the game.
+
+    This function increases the round and turn counters, requests a move
+    and returns a new game_state.
+
+    Raises
+    ------
+    ValueError
+        If gamestate['gameover'] is True
+    """
+    # TODO: Return a copy of the game_state
+
     # if the game is already over, we return a value error
     if game_state['gameover']:
         raise ValueError("Game is already over!")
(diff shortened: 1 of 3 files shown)


Debilski:
BTW there is a new nice feature for debugging tests. Run them as `pytest --log-level info` for example and you will get something like this (last 5 lines):

        ____________________________________ test_finished_when_no_food[3] _____________________________________

        bot_to_move = 3

            @pytest.mark.parametrize('bot_to_move', [0, 1, 2, 3])
            def test_finished_when_no_food(bot_to_move):
                """ Test that the game is over when a team has eaten its food. """
                l = """
                ########
                #  0.2 #
                # 3.1  #
                ########
                """
                bot_turn = bot_to_move // 2
                team_to_move = bot_to_move % 2
                def move(bot, s):
                    if team_to_move == 0 and bot.is_blue and bot_turn == bot.bot_turn:
                        return (4, 1), s
                        # eat the food between 0 and 2
                    if team_to_move == 1 and (not bot.is_blue) and bot_turn == bot.bot_turn:
                        # eat the food between 3 and 1
                        return (3, 2), s
                    return bot.position, s
            
                l = layout.parse_layout(l)
                final_state = run_game([move, move], layout_dict=l, max_rounds=20)
        >       assert final_state['round'] == 1
        E       assert 0 == 1

        test/test_game.py:460: AssertionError
        ------------------------------------------ Captured log call -------------------------------------------
        game.py                    536 INFO     Bot 0 moves to (3, 1).
        game.py                    536 INFO     Bot 1 moves to (4, 2).
        game.py                    536 INFO     Bot 2 moves to (5, 1).
        game.py                    536 INFO     Bot 3 moves to (3, 2).
        game.py                    548 INFO     Bot 3 eats food at (3, 2).
        ======================================= 4 failed in 0.34 seconds =======================================

Debilski:
Merging. This also adds a spec for #487. The logic is now:

A bot can only be killed, when it is its own turn (ie. suicide) or in the same turn that another bot *actively* moves onto it. *Actively* excludes respawning but includes standing still. This means that a bot that sits on the respawn position of its enemy will not be killed automatically and in fact may still have a chance of rescuing itself, if its turn happens to come before the turn of the enemy (or if the enemy itself moves away).

Tests for this are added in 4232e70704bc24f35599bf4733e296efa2f08a49

PR #576, opened 2019-04-30
Adapt the Tk viewer to the new API and add it to run_game

### LINKED ISSUE #508: Tk with --stop-at 0 does not show the team names
Could be fixed by sending `set_initial` twice to the viewers (once before and after the bots have received their `set_initial` but that leads to Tk misbehaving.
Suggestion would be to send a pre-set_initial command to the viewers in a future API refinement.

Debilski:
This PR changes the Tk viewer to run with a (still unoptimised) `game_state` and gets rid of all the old datamodel/universe API. Not all features are ported yet and there may be silent failures for some of the changed keys in `game_state`. Also, there is an UI bug (on OS X at least) where the maze loses some of its elements during the game. It is yet unclear to me why this is.

Additionally, viewers are now initialised in `run_game` (or rather `setup_game`). To ensure that the viewer can properly control the running game (and also ensure that the game waits until the viewer is ready) and exit it correctly, some minimal control code has been added to the loop in `run_game`. It is our aim to keep this control code and the loop as simple as possible, so that users can build their own game loops easily for testing.

Fixes #508.

### DIFF
--- a/pelita/exceptions.py
+++ b/pelita/exceptions.py
@@ -1,0 +1,15 @@
+
+class FatalException(Exception): # TODO rename to FatalGameException etc
+    pass
+
+class NonFatalException(Exception):
+    pass
+
+class PlayerTimeout(NonFatalException):
+    pass
+
+class PlayerDisconnected(FatalException):
+    # unsure, if PlayerDisconnected should be fatal in the sense of that the team loses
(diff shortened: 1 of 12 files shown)


Debilski:
Closes #508.

PR #578, opened 2019-05-22
Improved error propagation when a client errors

Debilski:
Needs still some polishing and more tests (one assertion fails because game does not return a draw when both teams fail fatally).


Debilski:
Can be merged now but we should think about the way (and data structure) we propagate errors from a client so that the UI can show more info about what happened.

PR #601, opened 2019-06-18
RF: Change game_state['error'] to list of dicts.

### LINKED ISSUE #600: [discussion] Is it possible to have more than one error per (round, turn)?
If not, then I would suggest to change the error list to a dict with key (round, turn) as it would make checking for errors in the current round simpler. (And an error in the current round would always give you a random move.) Right now, we have a bug that a timeout gives two errors.
I would like to make it a rule that there can be at most one error per turn.

Debilski:
There is only one error per (round, turn) allowed and all lead to the
same outcome -> a random move. Mirroring this in the data structure
simplifies the checks.

Currently, the turn is in range(0, 3), ie. global, even though
the error state is split between the two teams and therefore we could
also use the team-internal turn.

Additionally, this fixes the double-error problem when a timeout occurred
and a user would get an additional error for making a wrong move.

Fixes #600.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -316,7 +316,7 @@ def setup_game(team_specs, *, layout_dict, max_rounds=300, layout_name="", seed=
         layout_name=None,
         team_names=[None] * 2,
         fatal_errors=[[], []],
-        errors=[[], []],
+        errors=[{}, {}],
         whowins=None,
         rnd=Random(seed),
         viewers=[],
@@ -519,6 +519,7 @@ def play_turn(game_state):
     game_state.update(update_round_counter(game_state))
 
     turn = game_state['turn']
+    round = game_state['round']
     team = turn % 2
     # request a new move from the current team
     try:
@@ -552,11 +553,9 @@ def play_turn(game_state):
         # are collected and added to team_errors
         exception_event = {
             'type': e.__class__.__name__,
-            'description': str(e),
-            'turn': game_state['turn'],
-            'round': game_state['round'],
+            'description': str(e)
         }
-        game_state['errors'][team].append(exception_event)
+        game_state['errors'][team][(round, turn)] = exception_event
         position = None
         game_print(turn, f"{type(e).__name__}: {e}")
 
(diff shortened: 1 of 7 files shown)


pelita/game.py
```diff
@@ -624,25 +623,34 @@ def apply_move(gamestate, bot_position):
...
-        bot_position = legal_moves[gamestate['rnd'].randint(0, len(legal_moves)-1)]
-        error_dict = {
-            "turn": turn,
-            "round": n_round,
-            "reason": 'illegal move',
-            "bot_position": bot_position
-            }
-        game_print(turn, f"Illegal move {bad_bot_position} not in {sorted(legal_moves)}. Choosing a random move instead: {bot_position}")
-        team_errors.append(error_dict)
+
```
otizonaizit: could you in this PR also rename `get_legal_moves` to `get_legal_positions`: just to be consistent everywhere with the terminology

pelita/game.py
```diff
@@ -624,25 +623,34 @@ def apply_move(gamestate, bot_position):
...
+    # unless we have already made an error, check if we made a legal move
+    if not (n_round, turn) in team_errors:
+        if bot_position not in legal_moves:
```
otizonaizit: Like above, this should become `in legal_positions` not `in legal_moves`

pelita/game.py
```diff
@@ -624,25 +623,34 @@ def apply_move(gamestate, bot_position):
...
+            error_dict = {
+                "reason": 'illegal move',
+                "bot_position": bot_position
+                }
+            # add the error to the team’s errors
+            game_print(turn, f"Illegal move {bot_position} not in {sorted(legal_moves)}.")
```
otizonaizit: The error message should say `Illegal position` not `Illegal move`, and the list should come from `legal_positions` not `legal_moves`

pelita/game.py
```diff
@@ -624,25 +623,34 @@ def apply_move(gamestate, bot_position):
...
 
     # only execute move if errors not exceeded
     gamestate.update(check_gameover(gamestate))
     if gamestate['gameover']:
         return gamestate
 
+    # Now check if we must make a random move
+    if (n_round, turn) in team_errors:
+        # There was an error for this round and turn
+        # (but not enough that the game is over)
```
otizonaizit: Maybe this comment should be rephrased as `(but no error was fatal)` so that it is clear what `not enough` means.

pelita/game.py
```diff
@@ -624,25 +623,34 @@ def apply_move(gamestate, bot_position):
...
+        # We execute a random move
+        bot_position = gamestate['rnd'].choice(legal_moves)
+        game_print(turn, f"Choosing a random move instead: {bot_position}")
```
otizonaizit: as above, `position` instead of `move`. No, actually even better would be:
`f"Setting a random position instead: {bot_position}"`

pelita/game.py
Debilski: Hmm, but a fatal error is different. I can just leave it out. ‘There was an error for this round and turn but the game is not over’

Debilski:
I somewhat prefer move in some of the examples, even though it is a position, because a move implies that it is a neighbouring position. ’Setting a random position’ means to me that it could be anywhere on the maze.

otizonaizit:
>because a move implies that it is a neighbouring position.
>’Setting a random position’ means to me that it could be anywhere on
>the maze.

'Setting a legal position at random'. 
This way it recalls the name of the method on bot, which is nice.

Debilski:
Hmm, ok. But if we start renaming move → position. Where do we stop? What about things like `apply_move` (which takes a position as an argument)?

otizonaizit:
We'll rename as we go and where it makes sense. "apply_move" refers to applying the value obtained from the "move" function, so the name describes the intent correctly. The  "move" function refers to "moving" the bot, which is realized by returning a new position.
Everything makes sense.

Debilski:
I agree with this. (Changed some parts but there are sill a few variables left in the tests I think.) 

But in my mind, I think of ‘move’ as ‘move to position’, so it is almost always equivalent to me. Maybe we could get rid of the confusion, if we rename the remaining `Bot.legal_moves` to `Bot.legal_directions` and have move *always* mean ‘neighbouring position’ and not direction.

otizonaizit:
Yes, good idea. As a guide for intuition, think of "move" as verb, i.e. an action, and "position" as an object, i.e. the result of an action.

Debilski:
Good to go from my side. Let’s do `Bot.legal_moves` separately.

PR #610, opened 2019-07-14
Renamed --dump and added storage option for a player’s stdout/stderr

Debilski:
`--dump` is now called `--write-replay`

`--store-output` will write the player’s stdout and stderr into the specified folder. Note that this needs to be done unbuffered or else we will lose the most important messages (the reason why a bot crashed).


Debilski:
Still needs a fix for the tournament and we need a way to prefix the files written by store-output (or the tournament needs to create a subdirectory for each run).

Debilski:
Fixed the tournament part.

PR #611, opened 2019-07-18
RF: Remove dataclass for Python 3.6 compability.

Debilski:
Conda regularly defaults to Python 3.6, depending on the package that one installs, so I think it is still useful to be Python 3.6 compatible.

That said, I still prefer writing out the game state attributes in a class-like structure, as it makes documenting the attributes and their types much more obvious. However, doing this manually without the help of `@dataclass` results in unneeded (and brittle) code duplication, so we should probably keep it like this.

Not sure, if it is useful to make this a separate function `make_gamestate` since it will also result in much duplication. This means, that `setup_game` will be the *only* canonical way to create a game state.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -1,6 +1,5 @@
 """This is the game module. Written in 2019 in Born by Carlos and Lisa."""
 
-import dataclasses
 import itertools
 import logging
 import os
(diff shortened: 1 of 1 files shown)


Debilski:
Corollary: Travis can and should be trusted again before merging PRs.

PR #618, opened 2019-07-22
Monster PR with all the sprint work

otizonaizit:
@Debilski : here for you comments ;)


Debilski:
Will start reviewing after lunch. :)

otizonaizit:
Please have a look at 740c5eb0dc95a6c573e800f5433d58238cc610c3 .
That one is particularly intrusive.

Debilski:
Do you mind me merging my PRs (#602, #611, #612, maybe #614) to core-2.0 before and then rebasing your branch on top to make it a little smaller?

Debilski:
Ah, the raising of player exceptions is something I’d also considered already. I am not sure it is needed for NonFatalExceptions though. The only non-fatal exception is a timeout (which only occurs in a remote game anyway) and is not exactly the same thing.

An alternative to adding a flag and cluttering the exception handling could be to add a helper function to the tests `check_no_exception(game_state)` which would raise if a player has a fatal_error. I’ll think a little about the pros and cons of each.

otizonaizit:
But an illegal position is also an exception, which can happen also in local/test games.

otizonaizit:
We are still working on the branch, so reading it doesn't sound ok. Especially because we have also three clones around. Maybe do the rebasing somewhere else just for the comments? I don't expect that much of a discussion anyway. Stuff you would consider a problem can be handled with standard issues...

Debilski:
An illegal position is currently fatal though … And a position that is valid but on a wall does not raise any exception at all but is just added to the list of errors inside play_turn …

I’d prefer that only exceptions that are raised inside `move` are forwarded but everything else like timeouts of wall moves are ignored by the flag.

Debilski:
Ok, you do finishing your work on the sprint branch and I’ll handle rebasing it to a new PR branch.

Issue #626, opened 2019-07-23
Track the number of food pellets that a bot has eaten


Debilski:

otizonaizit:
I am working on this. Should finish tomorrow.

Debilski:
Left as an exercise for the user.

PR #627, opened 2019-07-23
WIP: Food eaten

otizonaizit:
Introduce a counter of the number of food pellets eaten by each bot. Make this counter available to the `Bot` class in the client.

WIP: tests for the new feature are missing. Old tests are fixed. DO NOT MERGE!

Fixes: #626


otizonaizit:
I could add the tests here, but I thought over it and I think the whole thing is not a good idea. We are blowing up the API for something as easy as:
```python
def move(bot, state):
    ...
    food_eaten = 30 - len(bot.enemy[0].food)
```
as opposed to `bot.kills/bot.deaths/bot.eaten` which are non trivial to implement within a `move` function without doing mistakes (just look at the corresponding tests to see how involved that is), the `food_eaten` attribute is easily, and without mistakes, implemented. True, this only tells you how much food your team has eaten, and not the food the current bot has eaten, but it can still be extended if needed by keeping track of when you step on enemy food. Also easy to implement.
I think we can drop this PR and forget about it. @Debilski : agree?

Debilski:
I am not overly attached to it, but if we are keeping track of individual kills (and especially, if we want our teams to know the individual kills of the enemy), I thought it makes sense too. Maybe we can add a semi-external statistics module for this in the future. :)

otizonaizit:
Again, I am not talking about the fact that it doesn't make sense to keep track of the food eaten. I am saying that we want to avoid the same mistakes we were making in the past, i.e. bloating the API for one-liners functions. The kills are hard to keep track of, so we offer the utility attributes. The food is super easy to keep track of, so there's no need to offer anything. That's my reasoning.

Debilski:
> just look at the corresponding tests to see how involved that is

Keeping track of one’s individual death counts was similarly trivial with the old bot_respawned semantics though …

I’m only arguing whether we should tell the teams the food count of the enemy bots, since we are already telling the teams the kill and death count of the individual enemy bots. I don’t see how one is more relevant than the other. (The food count of individual bots can even be more useful since it enables you to track your opponents much better.) Other than that I don’t care at all about this.

(Figuring out one’s own kill count is the same one-liner as figuring out the food one has eaten. For a slim API we don’t need any of that.)

otizonaizit:
OK, then show me the one-liner to keep track of the fact that you have been killed in the previous round. And then the one-liner that counts how many kills and how many deaths your bot and your team has in total. Without using private attributes of the Bot object nor the newly introduced 'eaten', 'kills', and 'deaths' attributes. If you have this one-liners, we can add them to the docs as well as the one-liner to keep track of the food you have eaten, and get rid of 'eaten', 'kills', and 'deaths'. The previous 'respawned' is *not* a one-liner, because bot has to check if other.respawned is true, and keep that info into the global state if the dead bot needs to do anything in case it was respawned when it's its turn.

Debilski:
*Disclaimer*: I am not arguing that we should advise this, just stating that it is possible.

    bot_eaten = (len(bot.track) == 1) # and not bot.round == 1
    bot_other_eaten = (len(bot.other.track) == 1) # and not bot.other.round == 1

    food_eaten = initial_enemyfood - len(bot.enemy[0].food)
    enemy_food_eaten = initial_food - len(bot.food)
    # initial_enemyfood would be in a state somewhere or hardcoded …
    
    total_kills = (bot.score - food_eaten) // 5
    total_deaths = (bot.enemy[0].score - enemy_food_eaten) // 5

Things are nicer to set up with a state, however. If you want to track the individual kills that you did and track the individual food eaten, then one could figure this out just before returning from `move` (because you can predict that you are going to score). For the individual deaths, one might either update the counter when one’s bot_eaten is True (and ignore the other bot until it is it’s turn) or do a little more complicated dance with global state.

otizonaizit:
Come on, the bot.track trick is only working by using a side effect (which does not work if you get eaten before you get a chance to move, which is impossible in the tournament, but well), total_kills and total_deaths are also using a dirty trick, which would take a while to figure out, which is not direct at all, etc. And their are not one-liners, because the total kills and deaths are depending on the food count.

I am closing this PR.

PR #628, opened 2019-07-23
ENH: Exit remote teams when the game is over.

### LINKED ISSUE #621: check_gameover should send an exit message to the Teams
Otherwise running games within IPython will never close the subprocesses to `pelita-player` since all previous game states are going to be stored in IPython’s `Out` list.

Debilski:
Closes #621

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -319,6 +319,9 @@ def setup_game(team_specs, *, layout_dict, max_rounds=300, layout_name="", seed=
     # Send updated game state with team names to the viewers
     update_viewers(game_state)
 
+    # exit remote teams in case we are game over
+    check_exit_remote_teams(game_state)
+
     return game_state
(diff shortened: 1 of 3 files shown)


Debilski:
There is still a problem when `run_game` is used with a controller and `controller_exit` is triggered (i.e. when someone closes the Tk window). In this case, we never set `gameover` and also never go to a path that would run `check_exit_remote_teams`. [*] As `run_game` is usually not used with a controller (at least not excessively), this should not be a big problem in practice, but it is something that we should watch.

This is not a problem when `call_pelita` is used, as `call_pelita` will (hopefully) take care of all the cleanup, so `call_pelita` is still the safer way to run pelita in some situations.

[*] I did not want to clutter the code any further and also it is not clear that exiting is always the correct thing to do.

Issue #624, opened 2019-07-23
Add traffic light to Tk viewer to show whose turn it is


Debilski:
Something in the spirit of

![](https://user-images.githubusercontent.com/216179/61712965-c592c100-ad57-11e9-937b-db6b9b3ed2b3.gif)

A more advanced design could even do a radial wipe animation inside the full circle for the three second count down. :)

Debilski:
Demo video … https://vimeo.com/349846095

otizonaizit:
if the code is not too terrible, make a PR of this! it's not bad at all.

otizonaizit:
when the grid is on this should have the bot_idx on top, or right below the corresponding lamp.

Debilski:
I think it should be a little smaller, I find it still too irritating. But then the lack of anti-aliasing kicks in and it looks pretty bad again …

otizonaizit:
It's good enough. PR!

PR #630, opened 2019-07-24
ENH: Add traffic lights to Tk to show current bot.

Debilski:
Initial version for #624 .


Debilski:
Alternate version in second commit. I like this one better. Screenshot from 2019-07-24 17-20-33

Debilski:
Or into the corner maybe. But with four elements on top of each other, it is too crowded. Might work, if we put the layout name to the left.
image

Debilski:
Closes #624

PR #634, opened 2019-07-31
WIP: TK-ui cleanups

otizonaizit:
- makes subtitle status line font smaller and more spaced from title (now the two lines should not overlap anymore (needs to be tested on anaconda and Mac)
- still some other cleanups here

### DIFF
--- a/pelita/ui/tk_canvas.py
+++ b/pelita/ui/tk_canvas.py
@@ -148,7 +148,7 @@ class TkApplication:
 
         self.ui = UI()
 
-        self.ui.header_canvas = tkinter.Canvas(master, height=38)
+        self.ui.header_canvas = tkinter.Canvas(master, height=50)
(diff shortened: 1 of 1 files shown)


otizonaizit:
Looks good with anaconda on Linux too. Any Mac to test?

Debilski:
> Any Mac to test?

Still trying to figure out why Tk breaks (on) all my systems.

Debilski:
On Anaconda macOS it is quite spacious now. But seems ok. Can be merged.

otizonaizit:
@Debilski : maybe we should switch off the codecov/patch check on PRs? If it would be possible to have a soft-fail, like a yellow circle, than OK. But a PR like this seems broken for no reason.
Also, could you merge this one too once pelita-core-2 is in? Thanks!

PR #636, opened 2019-08-02
ENH: do not pollute /tmp when creating communication files

otizonaizit:


Debilski:
Lgtm.

otizonaizit:
nope. it breaks the tests, and it shouldn't.

PR #642, opened 2019-08-03
remove layout with dead-end from list of layouts without dead-ends

otizonaizit:


Debilski:
Probably useful to decide with or after #646

otizonaizit:
Superseeded by #649

PR #646, opened 2019-08-05
BLD/RF: Add layouts as package_data, remove compressed __layouts.py

Debilski:
Moved all layouts to `pelita/_layouts` (and added an `__init__.py` to this folder). This PR should be thoroughly tested before we merge it. Unfortunately, having the layouts in the file system only, somewhat reduces the possibility to use tagging features in order to select the proper one. Also `pkg_resources` means that we cannot use `pathlib` for traversal and `resource_string` returns a byte object that we have to convert to `str` before we can use it.

Other than that, the features are hopefully the same.


Debilski:
Not sure, if I am doing it right, but this seems to work:

    python setup.py bdist --format=zip
    PYTHONPATH=dist/pelita-2.0.macosx-10.7-x86_64.zip python -m pelita.scripts.pelita_main

otizonaizit:
can be merged?

Debilski:
Short test on Windows throws no errors, so yeah.

PR #647, opened 2019-08-05
ENH: Only transmit the walls during set_initial phase

### LINKED ISSUE #485: Only send the maze in the set_initial phase (network-only)
Could be implemented transparently by patching `RemoteTeamPlayer` (on the sending site) and `SimpleClient` (on the receiving end). But it would need some kind of deep packet inspection which feels somewhat evil to do.

Debilski:
Closes #485.

The speed increase is not super big (10–20% faster), so I am not 100% sure, if we really do need it. However, compared to pelita from September 2018, we are almost three times slower, when running a game with stopping players on a big layout… At some point we should see, if profiling helps (or cache the height/width of the layout, which I am almost sure is a factor).

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -391,7 +391,9 @@ def prepare_bot_state(game_state, idx=None):
 
     """
 
-    if game_state.get('turn') is None and idx is not None:
+    bot_initialization = game_state.get('turn') is None and idx is not None
+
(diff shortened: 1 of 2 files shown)


Debilski:
Tests are still broken …

Debilski:
Fixed in 104bc825

PR #649, opened 2019-08-06
WIP: detect food chamber in existing layouts and remove them while generating them.

otizonaizit:
Many of our layouts contain food chambers, i.e. rooms in the maze that can be entered only through a single square, that contain food. Given that finding them is relatively easy with networkx and that sitting on the entrance is a winning strategy, we decided to get rid of food chambers from the default layouts.
In the screenshot the selected square indicates the entrance to a food chamber:

screenshot3

I have added one script that detects food chambers in existing layouts: `detect_dead_ends.py`. Run it like this

```bash
python detect_dead_ends.py layoutfile
```

Will show some output when a food chamber is found. One layout with a food chamber is for example: `normal_without_dead_ends_003.layout`.

I also modifiex `maze_generator.py` to remove food chambers while creating new mazes.

Open problems:

  - both scripts are messy and must be improved for readability
  - the porting of the old `maze_generator.py` from the old version of `networkx` to the new must have introduced some errors in the removal of the dead ends (proper, not the chambers), because now almost every layout generated has multiple chambers, while it was happening maybe 10%/15% of the time before. I suspect that the `remove_all_dead_ends` function in `maze_generator.py` has been broken by the port. On top of that, I think our graph is not directed?
  - As it stands now, we can detect food chambers in old layouts, and this would open the possibility to repair them by hand (the detect script will catch errors if the resulting maze is not symmetric, for example), but
  - I'd very much prefer to be able to generate new mazes that already follow the new specs.

What does everybody think?

Typical workflow when working with this branch:
```bash
$ cd pelita/layouts
# create new layout
$ python create_maze.py --seed 1 maze_generator.py > test.layout
# test that it works
$ python detect_dead_ends.py test.layout
# watch it in its full glory in pelita
$ pelita --layoutfile test.layout --stop-at 0
```


Debilski:
One remark: I wonder if we really need networkx for this and if it would not be easier (and more obvious) to do it all in numpy. (The bfs_successors and node.degrees is easily written and we might also find something for the connectivity.)

otizonaizit:
I think the comment is beside the point of the PR discussion, but well, we decided to ditch our own home-grown Graph object in favour of the more mature and feature-reach networkx-based implementation, and you propose to recreate it again  this time based on bumpy?

Debilski:
It was just something I noticed when looking at the code myself and trying to adapt it to numpy 2 properly. There are a few assumptions and constraints in the way the graph is generated that are easy and natural to do in numpy but much harder in networkx. Ie. networkx allows any connection between two nodes, which means you will have to do proper distance checks when you try to pinch trough a wall. (And networkx being agnostic about the 2d structure cannot even help you to find the best point to pinch trough, as it does not know anything about which connections are possible.) I just think it is mostly the wrong tool for the job here.

Of course, once the maze is ready, I’d prefer to use networkx for everything.

Debilski:
Why do we need a directed graph btw?

otizonaizit:
@Debilski @pberkes : I found the BUG!!! Just look at 7edacd897a34f1aea7f905e3d296d67e1a8121ab

After all, Pietro, it wasn't your fault that the dead_end removal code was broken since 2015!

otizonaizit:
Note that the CI failure is unrelated to this PR:
```
_______________________________ test_call_pelita _______________________________
    def test_call_pelita():
        rounds = 200
        viewer = 'ascii'
        filter = 'small'
        teams = ["pelita/player/StoppingPlayer", "pelita/player/StoppingPlayer"]
        (state, stdout, stderr) = call_pelita(teams, rounds=rounds, viewer='null', filter=filter, seed=None)
        assert state['gameover'] is True
        assert state['whowins'] == 2
        # Quick assert that there is text in stdout
        assert len(stdout.split('\n')) > 0
>       assert "Finished" in stdout
E       assert 'Finished' in "Replay this game with --seed 8417015225840030992\nUsing layout 'small_with_dead_ends_010'\n\x1b[94mᗧ\x1b[0m blue team 'pelita/player/StoppingPlayer' -> 'Stopping'\n\x1b[91mᗧ\x1b[0m red team 'pelita/player/StoppingPlayer' -> 'Stopping'\n"

test/test_libpelita.py:32: AssertionError
```

@Debilski : why is that test failing randomly?

Debilski:
Well, shame on me. I’m still trying to figure out why the networkx version would not visit all nodes (and why mdp would). But 7edacd8 also tells us what the original purpose of that `def collect_dead_end` was.

Debilski:
The failing test tells us that there might still be an issue with flushing to stdout. :(

Debilski:
(I restarted travis)

otizonaizit:
I am in the process of refactoring the whole thing. We got rid of `DiGraph`, and I added a bunch of tests for the dead-end removing code. Still to do is the refactoring of the horrible food-chamber detection and clean up code. Will do that next.

Debilski:
Regarding 00d0a2e: I wouldn’t put the scripts into the same directory as the layouts. It makes it harder to see what scripts we have.

otizonaizit:
Yes, that's the plan. For now I kept the scripts there because it makes testing on the terminal easier, but the scripts will be moved out soon.

otizonaizit:
@Debilski : I think I am done with this PR. What changed:

- mazes are now generated without dead ends and without chambers (rooms with only one door), i.e. we do not have a way to generate mazes with dead ends and/or chambers anymore
- the maze generation stuff is now thoroughly tested (also some reasonable edge cases are tested, for your pleasure)!
- I promoted the maze generation functions to their own module in the main pelita namespace
- the script to generate one new random layout is now installed by default and can be called from anywhere
- the `_layouts` directory now only contains the old layout files

Nice to have for the future, but not done now: get rid of the `numpy` dependency. It is not really needed in the maze generation, I think the whole thing can be rewritten using the default layout parsing that we already use in the rest of pelita which returns a dictionary.

For the next PR, to be done very soon once this is merged, I plan to regenerate all layouts, note that I changed the names, now they are called `size_NNN.layout`, where `size` can be `small`, `normal` or `big`. This will require some adjusting in the rest of pelita, but I can take care of it.
Only problem for now: generating a big (64x32) layout takes 7 minutes on my laptop, so I am not sure it is worth to waste CPU-cycles *and* storing space on installation for layouts that are not used anyway. If someone wants to try one or two, they can generate them on the fly with the script. I don't see the point of installing 100 of them. **And:** playing 300 rounds on a big layout with stopping player and without a viewer takes 4.3sec on my laptop, on a normal layout 2.3sec and on a small one 1.5 sec. Not sure why it scales that way? 

How does this sound? Merge?

Debilski:
Looks good. But I want to spend another look at it in daylight. There must be a way to speed things up.

Half-related: Do we actually care about dead ends when there is no food inside?

> Not sure why it scales that way?

I think it’s the noiser. Amongst others our check `pos in walls`. Admittedly, having a list of wall position tuples isn’t something that scales very well (especially if Python doesn’t care that we have sorted it).

Debilski:
I think there is one edge-case bug: Infinite loop when more food is specified than would fit.

otizonaizit:
> I think there is one edge-case bug: Infinite loop when more food is specified than would fit.

fixed in ad9c8a2b25b43a09adfe3c731eceaa060f0b9efd

otizonaizit:
> There must be a way to speed things up.

Maybe, but it is not worth it now. This is a one-time cost.

> Half-related: Do we actually care about dead ends (chambers) when there is no food inside?

Yes, because they would be a portion of the maze where no one would have any reason to enter, making the maze effectively smaller.
 
> > Not sure why it scales that way?
> 
> I think it’s the noiser. Amongst others our check `pos in walls`. Admittedly, having a list of wall position tuples isn’t something that scales very well (especially if Python doesn’t care that we have sorted it).

Hmm, I don't buy it. Our lists of tuples are super small. But if `in` is the problem, we should use a set. That should be super fast, and our list of coordinates are inherently a set.

Debilski:
> > I think there is one edge-case bug: Infinite loop when more food is specified than would fit.
> 
> fixed in [ad9c8a2](https://github.com/ASPP/pelita/commit/ad9c8a2b25b43a09adfe3c731eceaa060f0b9efd)

    pelita-createlayout -y 6 -x 8 -f 10

Debilski:
> > There must be a way to speed things up.
> 
> Maybe, but it is not worth it now. This is a one-time cost.

Yeah, but 7 minutes … (Or 1 minute for regular mazes …) We should definitely also create *a few* of the big mazes. No need for 100. But 10 maybe.

> 
> > Half-related: Do we actually care about dead ends (chambers) when there is no food inside?
> 
> Yes, because they would be a portion of the maze where no one would have any reason to enter, making the maze effectively smaller.

Good point.

> 
> > > Not sure why it scales that way?
> > 
> > 
> > I think it’s the noiser. Amongst others our check `pos in walls`. Admittedly, having a list of wall position tuples isn’t something that scales very well (especially if Python doesn’t care that we have sorted it).
> 
> Hmm, I don't buy it. Our lists of tuples are super small. But if `in` is the problem, we should use a set. That should be super fast, and our list of coordinates are inherently a set.

I mean, I profiled it. It is in part the noiser. The algorithm does a little bit too much work currently, which is why it is doing the `in` check far too often. So there is potential for optimisation. Additionally, we spend a lot of time in the `ensure_tuples` functions in pelita-player. Still, there are parts of it that can be optimised easily but in general we will have to admit that pelita-core-2.0 is somewhat inherently slower than pelita 1.0 due to its unoptimised datastructures.

otizonaizit:
> > > I think there is one edge-case bug: Infinite loop when more food is specified than would fit.
> > 
> > 
> > fixed in [ad9c8a2](https://github.com/ASPP/pelita/commit/ad9c8a2b25b43a09adfe3c731eceaa060f0b9efd)
> 
> ```
> pelita-createlayout -y 6 -x 8 -f 10
> ```

Thanks. I took the chance and refactored `add_food` so that is more robust now and so that it fails gracefully. Tests are also there: bc1dca6293718891ed875fcc3cc7b2e515019295

otizonaizit:
> > > There must be a way to speed things up.
> >
> > Maybe, but it is not worth it now. This is a one-time cost.
>
> Yeah, but 7 minutes … (Or 1 minute for regular mazes …) We should definitely also create _a few_ of the big mazes. No need for 100. But 10 maybe.

I can start the creation offline right now. Tomorrow or on Monday I can push the newly created layouts. With 10 big ones and 100 of the other kind
 
> > > > Not sure why it scales that way?
> > >
> > > I think it’s the noiser. Amongst others our check `pos in walls`. Admittedly, having a list of wall position tuples isn’t something that scales very well (especially if Python doesn’t care that we have sorted it).
> >
> > Hmm, I don't buy it. Our lists of tuples are super small. But if `in` is the problem, we should use a set. That should be super fast, and our list of coordinates are inherently a set.
> 
> I mean, I profiled it. It is in part the noiser. The algorithm does a little bit too much work currently, which is why it is doing the `in` check far too often. So there is potential for optimisation. Additionally, we spend a lot of time in the `ensure_tuples` functions in pelita-player. Still, there are parts of it that can be optimised easily but in general we will have to admit that pelita-core-2.0 is somewhat inherently slower than pelita 1.0 due to its unoptimised datastructures.

Would you be up to open an issue about this, with the procedure you used for profiling and the result of your profiling. We want to keep track of this.

otizonaizit:
Merge?

Debilski:
Yesyes, merge.

Debilski:
It is deliberate that there is no food on the border?

otizonaizit:
>It is deliberate that there is no food on the border?

Yes, that was in the description of the original function since the beginning. It makes sense to me. There was another thing, the food wouldn't be located within a radius of 6 from the pacmen starting positions. I changed that to be only exactly the starting positions that have no food (otherwise you could be killed by a respawned pacman). Otherwise the food can be anywhere. (The 6 was hard coded, so it would have also crashed when creating a smaller maze)

PR #650, opened 2019-08-07
WIP: Early draft for simulate_move.

Debilski:
Early draft for simulate_move so that it does not get lost (still a draft. DO NOT MERGE). This version one is capable of only doing a single step and returns the current bot object but with score and positions updated.
This is not yet enough, however, as the Bot object crucially is missing info on whether the game has finished (and who has won). (It makes sense that the Bot object does not include this data, as in a normal game situation, the game is by definition still not over.)
More elaborate simulation will also require that we can simulate more than one move (one will at least want to simulate all possible enemy movements in advance in order to know which positions to step on).

My suggestion for now is that I keep working on possible designs in my own player and we postpone this until we have found a good solution.

### DIFF
--- a/pelita/utils.py
+++ b/pelita/utils.py
@@ -26,6 +26,54 @@ def split_food(width, food):
     return team_food
 
 
+def bot_to_gamestate(bot):
+    if bot.is_blue:
+        turn = bot.turn * 2
+        teams = [bot, bot.enemy[0]]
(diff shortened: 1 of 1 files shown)


Debilski:
Closing for now. The whole situation of converting back and forth from Bot <=> game_state is really messy.

PR #653, opened 2019-08-08
Removing numpy dependency in tournament mode

Debilski:
Numpy was used to make the evaluation of the knock out mode a little nicer but we can do without. Removing numpy means we are free to remove numpy from our setup.py (even though we might still need it for the maze scripts), which in turn makes installation of pelita in a new environment a little more lightweight.

Is this worth going for? If so, I’ll add more comments to the numpy-free version and make it slightly clearer what is going on.


otizonaizit:
I don't think it's worth it. It is only a theoretical concern, you need `numpy` for everything. And for the maze generation we are going to need it in any case. And, I was thinking of adding an example of conversion of the maze to a numpy array in pelita_template, so -1 from me.

Debilski:
Ok, we keep it in the setup.py. But I think I’ll still remove it from the tournament. (I run pelita from a network home folder. Every time that numpy is imported costs time.)

lschwetlick:
>And, I was thinking of adding an example of conversion of the maze to a numpy array in pelita_template
I have a notebook that does this and plots the game state using matplotlib. If you think thats a good idea, I'll clean it up and push it to Pelita_Template. 
Screenshot 2019-08-08 at 18 39 10

Debilski:
I like it (if you manage to tune the colours)! If the code is clean and extensible enough, then the students can use it as a starting point for their own analyses. (Extending matplotlib is probably more sensible for our students than learning how to write SVG.)

otizonaizit:
That's a very good idea. Can you make it an utility function or a demo, i.e. not a notebook? The notebooks are very bad citizens in git... 
We could have an example notebook too show how to play around with pelita in the notebook, but useful functionality should be runnable without it, IMHO.

lschwetlick:
Well, I can easily just put the plot function and the convert function into a .py script and import it into the notebook as an example. Is that what you mean?

otizonaizit:
I mean, have a demo that has a function that converts the maze into the numpy array and uses the array in the bot to do something useful and write a test for the demo that tests the round-trip maze-numpy-maze. The plotting can be an additional function in the demo, but it doesn't need tests.

Than in the notebook you use these things to do whatever fancy things can be done.

I'm assuming that there are many ways to convert from maze to numpy, each one useful for different purposes. We shouldn't give the impression that there is a "canonical" conversion, or the students will be stuck with it.

lschwetlick:
There are two things here, the visualization and the numpy stuff. 
Concerning the visualization, to be honest, I feel like it might be worthwhile to have a notebook showing how to visualize individual board states independently of whether they actually want to use the numpy. I find it super useful to have the notebooks also to just play around with, especially in the beginning if the NB already has all the imports and game setup in place and ready to be used.

And about the numpy I'm kind of having trouble thinking of something useful for it to do that is also basic enough that we would want it in a demo. 
I never actually converted the whole maze to numpy (whatever that means), but rather made a matrix of the correct size and played with different factors to activate the fields. That isnt really intuitive enough to describe in a demo nor was it effective enough that I'd want to. Any ideas?

@Debilski : Is this visualization more aesthetically pleasing? 
Screenshot 2019-08-08 at 20 32 16

Debilski:
> @Debilski : Is this visualization more aesthetically pleasing?
> Screenshot 2019-08-08 at 20 32 16

Ah, I only meant that the food and the bots have their blue and red tones. (But just as in your first example, I think it’s a nice touch to have a slightly different colour for each bot, and maybe for the food too.) the food could be a little smaller, I think, but it’s something the students can and should adapt to their own liking, so better not make it too complicated.

It would be cool if the code makes it obvious how to add a new shape to a field on the maze, or how to change the background colour of a field. Then students can write code for example that shades the area of noise (ie. the potential positions of an enemy bot) in a different colour. (I think it would go too far, if we provided the exact code for this but it should be easy to figure out and implement.)

otizonaizit:
Merge?

Debilski:
Not yet, I am still working on #664 and doing a general cleanup of the tournament code.

Debilski:
I can also make a new PR for this though, so go ahead if you want to merge.

otizonaizit:
If I understand correctly, you moved `tournament.py` to `__init__.py`, but without using `git mv`, so git could not warn you of the conflicts. The result is that the `__init__.py` still uses the obsolete `--filter` option and does not contain the improvements to `call_pelita` that have been merged today. Can you fix it? 
Otherwise, it looks good to me.

Debilski:
The conflicts only appear when merging. But yeah, I should have rebased earlier.

Debilski:
Wow. That involved a lot of git magic to rebase …

otizonaizit:
yes, but that is because you did not `git mv` the file...

otizonaizit:
Should I review from scratch or can you tell me which commit is the one addressing the move from `tournament.py` to `__init__.py`?

Debilski:
There is no such thing as `git mv`. This is just `git rm` + `git add`. Which is what I did. The problem is that the `__init__.py` was already there and so git would not see it as a rename. I had to cheat it a little.

Debilski:
The move is in the last commit. (`git diff master:pelita/tournament/tournament.py HEAD:pelita/tournament/__init__.py`)

PR #655, opened 2019-08-08
Show noisy enemies with a questionmark

### LINKED ISSUE #622: print(bot) should show which enemies are noisy
This will hopefully avoid confusion.

One remark: since we got rid of set_initial in the new-style API, the teams never see their enemies sitting unnoised on their initial positions, which has been a nice (and easy) starting point for filtering. Question: Do we want to be explicit about how the initial positions are fixed (ie. add an example) or do we want them to figure it out themselves?

Debilski:
> print(bot)

    StoppingBots (you) vs Random Explorer Players.
    Playing on blue side. Current turn: 1. Round: 2, score: 0:0. timeouts: 0:0
    ################################
    #   ..   .     ##             ?#  
    #  ## ...#  . .   ##.## ###### #
    # . # . .#        .#. #. . #   #
    # . #### ########   . #  # ?   #
    #      .   .       #  #.## #. .#
    #   #######.##     #  # #  #. .#
    #  ... #    #  ##. #  #    #   #
    #   #    #  # .##  #    # ...  #
    #. .#  # #  #     ##.#######   #
    #. .# ##.#  #       .   .      #
    #     #  # .   ######## #### . #
    #   # . .# .#.        #. . # . #
    #0###### ##.##   . .  #... ##  #
    #1             ##     .   ..   #
    ################################

Not sure if this is legible enough?

Closes #622

### DIFF
--- a/pelita/layout.py
+++ b/pelita/layout.py
@@ -316,7 +316,7 @@ def parse_single_layout(layout_str, num_bots=4, allow_enemy_chars=False):
         out['enemy'] = enemy
     return out
 
-def layout_as_str(*, walls, food=None, bots=None, enemy=None):
+def layout_as_str(*, walls, food=None, bots=None, enemy=None, noisy=None):
     """Given walls, food and bots return a string layout representation
 
     Returns a combined layout string.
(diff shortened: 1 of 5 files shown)


otizonaizit:
and you would show `E` when the position is not noisy? This seems a good idea! Have you adapted also the layout_from_str function to understand the new character?

Debilski:
> On 8. Aug 2019, at 17:42, Tiziano Zito <notifications@github.com> wrote:
> 
> and you would show E when the position is not noisy? This seems a good idea! Have you adapted also the layout_from_str function to understand the new character?
> 

Not yet. (Should have made it a draft PR.)

Debilski:
`parse_layout` can parse the ? now but it will be the same as E. I don’t know if we want `setup_test_game` to automatically set the `is_noisy` flag, when there is a question mark?

otizonaizit:
Sure, set is_noisy.

PR #657, opened 2019-08-08
NF: Add utils function to load a builtin layout for a team

Debilski:
Adds a new helper function to `pelita.utils` that can be used to load a builtin layout in a way that it can be passed to `setup_test_game`:

    layout = load_builtin_layout("small_without_dead_ends_001", is_blue=True)
    setup_test_game(layout=layout, is_blue=True)

If this is the proper way to do it is still up for discussion.

### DIFF
--- a/pelita/utils.py
+++ b/pelita/utils.py
@@ -1,8 +1,9 @@
 import logging
 import random
 
-from .player.team import create_layout, make_bots
 from .graph import Graph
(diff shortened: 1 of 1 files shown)


Debilski:
CC @lschwetlick

Debilski:
Whoops. Forgot to `git add` my test …

PR #658, opened 2019-08-09
Better walls for TK UI

otizonaizit:
This PR increases the thickness of all walls in the TK UI. This way it is easier to see how many squares are there between wall tunnels, for example. 
Also, fill in squares of walls. Until now there is empty space within a square of walls, and it seems it would be reachable. If we go for this thickness, then maybe the color of the walls could be made a little lighter.

Two screenshots on two different layouts to appreciate the change.

Master branch (example 1):
master_branch1

This PR (example 1):
this_branch1

Master branch (example 2):
master_branch2

This PR (example 2):
this_branch2

### DIFF
--- a/pelita/ui/tk_sprites.py
+++ b/pelita/ui/tk_sprites.py
@@ -221,7 +221,7 @@ class Wall(TkSprite):
 
 
     def draw(self, canvas, game_state=None):
-        scale = (self.mesh.half_scale_x + self.mesh.half_scale_y) * 0.5
+        scale = (self.mesh.half_scale_x + self.mesh.half_scale_y) * 0.6
         if not ((0, 1) in self.wall_neighbors or
                 (1, 0) in self.wall_neighbors or
                 (0, -1) in self.wall_neighbors or
@@ -230,7 +230,7 @@ class Wall(TkSprite):
             # draw only a small dot.
             # TODO add diagonal lines
             canvas.create_line(self.screen((-0.3, 0)), self.screen((+0.3, 0)), fill=BROWN,
-                               width=0.8 * scale, tag=(self.tag, "wall"), capstyle="round")
+                               width=scale, tag=(self.tag, "wall"), capstyle="round")
         else:
             neighbours = [(-1, -1), (0, -1), (1, -1), (1, 0), (1, 1), (0, 1), (-1, 1), (-1, 0)]
             for dx in [-1, 0, 1]:
@@ -246,7 +246,8 @@ class Wall(TkSprite):
                             pass
                         else:
                             canvas.create_line(self.screen((0, 0)), self.screen((2*dx, 2*dy)), fill=BROWN,
-                                               width=0.8 * scale, tag=(self.tag, "wall"), capstyle="round")
+                                               width=scale, tag=(self.tag, "wall"), capstyle="round")
+
 
 class Food(TkSprite):
     @classmethod
--- a/pelita/ui/tk_sprites.py
+++ b/pelita/ui/tk_sprites.py
@@ -248,6 +248,13 @@ class Wall(TkSprite):
                             canvas.create_line(self.screen((0, 0)), self.screen((2*dx, 2*dy)), fill=BROWN,
                                                width=scale, tag=(self.tag, "wall"), capstyle="round")
 
+            # if we are drawing a closed square, fill in the internal part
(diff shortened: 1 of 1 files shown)


Debilski:
-0. Not sure what it is, but something is weird about that design.

I think if we want to fill it and have thick borders, maybe we should not have round corners.

otizonaizit:
by the way, the unrelated travis ci failure are annoying. Couldn't we just skip that test until is fixed?

otizonaizit:
> I think if we want to fill it and have thick borders, maybe we should not have round corners.

I like the round corners. Anyway, here is a proposal with less prominent walls: 1627de6
screenshot2

lschwetlick:
I prefer the black walls. The faded version looks like theres something wrong with my screen.

Debilski:
> I prefer the black walls. The faded version looks like theres something wrong with my screen.

Agreed. Maybe it would help to have a dark black small border around the faded wall?

otizonaizit:
> > I prefer the black walls. The faded version looks like theres something wrong with my screen.
> 
> Agreed. Maybe it would help to have a dark black small border around the faded wall?

Unfortunately the TK `create_line` method for the canvas does not allow for an outlline in the line
object. Either we redo the whole thing using polygons, or we choose one single fill color.
You both seem to prefer the old dark brown?
@lschwetlick : do you prefer the version in this PR with thicker walls and filled-in squares or the old version with thin walls and hollow squares? (see above for the screenshots)

otizonaizit:
Also, we are probably the worst to judge, as we are playing with this thing with a while...

Debilski:
> You both seem to prefer the old dark brown?

Not necessarily the brown, but I think it should have the same brightness as the bots and food.

> Also, we are probably the worst to judge, as we are playing with this thing since a while...

Not only that but we are also probably not the best people to design this properly in the first place …

otizonaizit:
> > You both seem to prefer the old dark brown?
> 
> Not necessarily the brown, but I think it should have the same brightness as the bots and food.
> 

What do you mean by "brightness"? You mean the V in HSV coordinates? Give the colors in master, we have for red (0,62,92) and for blue (209,57,85), so the two are already not equally "bright". The brown is (9,54,19). The grey in my last screeshot is (26,6,48).

What do you think of this?
screenshot1

This is halfway between my grey and the original brown in terms of brightness:
It is RGB (89, 86, 84) and HSV (26,6,35).
it seems better to me. It's not so eye-catching like the original brown, which is a bit too much if the walls are thicker, but it's a bit more visible than the gray in my previous example.

lschwetlick:
> do you prefer the version in this PR with thicker walls and filled-in squares or the old version with thin walls and hollow squares? (see above for the screenshots)

Hm, I think filled in and thicker is better... Maybe we can do a survey at the school :)

Debilski:
> What do you mean by "brightness"? You mean the V in HSV coordinates? 

Probably. I think I am not sure what the proper term is for that colour impression.

(Also: On my computer I never really noticed that the original was in fact _brown_ but rather simply dark to me.)

The new version is a little better but still, hmm, maybe we just need to get used to it? The shapes are not really symmetric though.

image

Debilski:
I think I prefer 5153d93 more than the new one. It is still too milky.

otizonaizit:
OK, merging the version that makes walls thicker and the squared filled in, but no changes in color

Debilski:
I think the filling algorithm can still be improved a little:

```
################################
#     #   #    ##  .      .   3#
#.###   # #        #..#..   . 1#
#  . .# # ##.#.   ### ###.######
#  #  # # #       .     .  .   #
# .# ##    . .    ####.#    ## #
#    #  # #  . ##    #.#    . .#
### ## .# ## #####..  .#..     #
#     ..#.  ..##### ## #. ## ###
#. .    #.#    ## .  # #  #    #
# ##    #.####    . .    ## #. #
#   .  .     .       # # #  #  #
######.### ###   .#.## # #. .  #
#0 .   ..#..#        # #   ###.#
#2   .      .  ##    #   #     #
################################
```

otizonaizit:
But that is an illegal layout anyway: it is a disconnected graph... To catch those you would need to either convert to graph or iteratively look for closed chambers. Not worth it in my opinion, because those layouts can not happen in practice.

Debilski:
It’s not disconnected. It just looks disconnected in Tk.

PR #661, opened 2019-08-12
Generate new layouts without chambers and change layout filtering

otizonaizit:
The built-in layouts have been regenerated using the new methods introduced by #649 .

The layouts have been renamed with the scheme `size_XXX.layout`. The layout loading functions and the pelita CLI have been adapated to use the new naming scheme. Also, the parameter to select the pool of layouts to use is now called `size` (it was confusingly called `filter` before).


otizonaizit:
by the way, may 2ecf920a53cd23b603e4f0557e1aa9eadf86fcad fix the random failures in test_call_pelita ?

Debilski:
> by the way, may [2ecf920](https://github.com/ASPP/pelita/commit/2ecf920a53cd23b603e4f0557e1aa9eadf86fcad) fix the random failures in test_call_pelita ?

If it works then it would only mask the problem in the test. The flushing needs to be done inside Pelita …

Debilski:
LGTM (as far as I can follow in the 900 file diff). Can we also have 10 tiny layouts (undocumented) for testing reasons? OTOH with a functioning pelita-createlaout we don’t really need to.

otizonaizit:
On 12 August 2019 15:59:51 CEST, Rike-Benjamin Schuppner <notifications@github.com> wrote:
>> by the way, may
>[2ecf920](https://github.com/ASPP/pelita/commit/2ecf920a53cd23b603e4f0557e1aa9eadf86fcad)
>fix the random failures in test_call_pelita ?>
>>
>If it works then it would only mask the problem in the test. The
>flushing needs to be done inside Pelita …>

Well, yes and no. I think there's no way you can guarantee that stdout will be flushed from a process. You can only guarantee that it will be flushed when someone tries to read from it. So, I don't think this is 100% fixable within pelita

otizonaizit:
Yep, they could be generated on the fly with a fixed seed. Which test do you want to feed them with? I can prepare the setup in this PR before merging.

Debilski:
> On 12 August 2019 15:59:51 CEST, Rike-Benjamin Schuppner ***@***.***> wrote: > by the way, may [[2ecf920](https://github.com/ASPP/pelita/commit/2ecf920a53cd23b603e4f0557e1aa9eadf86fcad)]([2ecf920](https://github.com/ASPP/pelita/commit/2ecf920a53cd23b603e4f0557e1aa9eadf86fcad)) fix the random failures in test_call_pelita ?> > If it works then it would only mask the problem in the test. The flushing needs to be done inside Pelita …>
> Well, yes and no. I think there's no way you can guarantee that stdout will be flushed from a process. You can only guarantee that it will be flushed when someone tries to read from it. So, I don't think this is 100% fixable within pelita

Yeah, but I think we can just fix it with an explicit flush in main pelita.

Also, I am not sure if your fix actually works. I think stdout will just be empty when the process exits before having flushed.

Debilski:
> Yep, they could be generated on the fly with a fixed seed. Which test do you want to feed them with? I can prepare the setup in this PR before merging.

Ah, I didn’t even think of unit tests. I’ll just use it on the command line. (I wanted to do some behaviour testing of teams killing each other. And with smaller mazes it simply happens more often, that was the whole idea.)

Debilski:
I think I’ve figured it out with the flushing (Hint: It’s not flushing). You can revert 2ecf920. Fix coming soon.

PR #663, opened 2019-08-12
ENH: Increase the chance that call_pelita flushes correctly to stdout.

### LINKED ISSUE #652: Stdout is not properly flushed at the end of a game.
_______________________________ test_call_pelita _______________________________
        def test_call_pelita():
            rounds = 200
            viewer = 'ascii'
            filter = 'small'
            teams = ["pelita/player/StoppingPlayer", "pelita/player/StoppingPlayer"]
            (state, stdout, stderr) = call_pelita(teams, rounds=rounds, viewer='null', filter=filter, seed=None)
            assert state['gameover'] is True
            assert state['whowins'] == 2
            # Quick assert that there is text in stdout
            assert len(stdout.split('\n')) > 0
    >       assert "Finished" in stdout
    E       assert 'Finished' in "Replay this game with --seed 8417015225840030992\nUsing layout 'small_with_dead_ends_010'\n\x1b[94mᗧ\x1b[0m blue team 'pelita/player/StoppingPlayer' -> 'Stopping'\n\x1b[91mᗧ\x1b[0m red team 'pelita/player/StoppingPlayer' -> 'Stopping'\n"
    
    test/test_libpelita.py:32: AssertionError

Debilski:
I think I have found our problem with the flushing. (Running `call_pelita` does not always return the last line in the game. Ie. ‘Finished’ is never printed.)

The problem is probably not flushing (we do in fact flush in `ResultPrinter` which prints the final line). The problem is that before we print the last line, we already tell `call_pelita` via zmq that the game is over. Therefore, `call_pelita` decides that it can as well exit and tries terminating out main pelita process. In most cases all is well and the main pelita process has already printed and exited. Sometimes it will be forcefully terminated before it can print.

This fix adds a small `wait(1)` after the `call_pelita` loop has received the message that the game has ended, which should be enough time for pelita to finish all its remaining jobs.

We might also just ensure that the zmq message is executed as the final thing before pelita exits but then again, this might also run us into bad situations when zmq is cleaned up while sending the message, so we will try it with the simple `wait(1)` for now.

Closes #652

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
 script:
 - >
     python -c "import zmq; print('Using pyzmq {} and zmq {}.'.format(zmq.pyzmq_version(), zmq.zmq_version()))" &&
-    echo "Sarting tests..." && python -m pytest test/ && echo "Tests finished." && echo "Starting pelita CLI as a module..." &&
+    echo "Starting tests..." && python -m pytest test/ && echo "Tests finished." && echo "Starting pelita CLI as a module..." &&
     python -m pelita.scripts.pelita_main --null && echo "Pelita CLI as a module finished." && echo "Starting pelita CLI as a script..." &&
     pelita --null --rounds 100 --size small $player && echo "Pelita CLI as a script finished."
 - >
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
 script:
 - >
     python -c "import zmq; print('Using pyzmq {} and zmq {}.'.format(zmq.pyzmq_version(), zmq.zmq_version()))" &&
-    python -m pytest --cov=pelita test/ &&
+    python -m pytest test/ &&
     python -m pelita.scripts.pelita_main --progress &&
     pelita --null --rounds 100 --size small $player 2>&1
 - >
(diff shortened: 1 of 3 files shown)


Debilski:
Fixed another randomly failing test in 7d5fda8 .

otizonaizit:
I was fixing the PR. Let me finish.

otizonaizit:
Sorry for force pushing, but that was the easiest option instead of opening a new PR.

otizonaizit:
Ready to merge, actually.

Debilski:
`str(seed)` is not stable. The calling site should ensure that it has passed a string. (Perhaps `call_pelita` should fail with a `ValueError`.)

otizonaizit:
what does it mean: it's not stable? call_pelita is not a user-facing function. we control what gets passed there, no?

otizonaizit:
On top of that, you pass round as a `int` too. I think there's no point in changing anything here. It's fine, it works.

Debilski:
> call_pelita is not a user-facing function. we control what gets passed there, no?

Exactly. I would have just fixed the failing test … Now you can pass it a numpy array as a seed but it will still break.

otizonaizit:
>>call_pelita is not a user-facing function. we control what gets passed there, no?
>
>Exactly. I would have just fixed the failing test … Now you can pass it a numpy array as a seed but it will still break.
>

But why then you allow ints for `round`? It is natural to pass ints for rounds and for seed. Why would we pass a numpy.array for the seed? If the function breaks in this case it's fine.

But I don't care, you can make the call_pelita signature as weird as you want. As long as the function is only used in the tournament code I don't care how it looks like.

PR #667, opened 2019-08-13
Get rid of our own graph implementation and use networkx instead

### LINKED ISSUE #666: Get rid completely of pelita.graph

otizonaizit:
Closes #666 : the number of the beast!


Debilski:
LGTM

PR #669, opened 2019-08-13
Refactor pelita CLI

otizonaizit:
Minor adjustements to the pelita CLI:

- unify the `--layout` and `--layoutfile` options
- update the example team specification to match the new API

### DIFF
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -171,7 +171,8 @@ def main():
         sys.exit(0)
 
     if args.list_layouts:
-        layouts = pelita.layout.get_available_layouts()
+        layouts = pelita.layout.get_available_layouts(size='all')
+        layouts.sort()
         print('\n'.join(layouts))
         sys.exit(0)
 
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -155,9 +155,8 @@ Team Specification:
     Demo players can be found at https://github.com/ASPP/pelita_template
 
 Layout specification:
-    If neither --layoutfile nor --layout are specified, the maze is
-    chosen at random from the pool of built-in normal-sized layouts.
-    You can change this pool by using --size.
+    If --layout is not specified, the maze is chosen at random from the pool of
+    built-in normal-sized layouts. You can change this pool by using --size.
 """
 
 
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -63,7 +63,7 @@ parser.add_argument('--replay', help=long_help('Replay a dumped game'),
 parser.add_argument('--store-output', help=long_help('Write all player’s stdout/stderr to the given folder (must exist)'),
                     metavar='FOLDER')
 parser.add_argument('--list-layouts', action='store_true',
-                    help='List all available layouts.')
+                    help='List all available built-in layouts.')
 parser.add_argument('--check-team', action="store_true",
                     help=long_help('Check that the team is valid (on first sight) and print its name.'))
 
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -136,16 +136,16 @@ Team Specification:
     * TEAM_NAME
         a string with the name of the team.
 
-    * move(turn, game) -> next_move
-        a function that takes the current game and returns the move for the bot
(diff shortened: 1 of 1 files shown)


Debilski:
I think the built-ins should have preference, no? While you’re at it: You could also combine the `--list-layouts` with `--size`

otizonaizit:
>I think the built-ins should have preference, no? While you’re at it: You could also combine the `--list-layouts` with `--size`

No, because then it becomes harder to run a local copy of a built-in layout that you just modified a little. So, no, if you call a file
small_001, then I am pretty sure you want to run that one and not the built-in small_001. Note that if you happen to name a random file like a built-in layout, pelita will fail gracefully.
Not sure what you mean by combining list-layouts and size? I don't see the point of complicating the help of list-layouts to allow you to select a certain size. You can grep for it. Or you meant something else?

Debilski:
If you just copy it over, it will have a suffix `.layout` anyway that you will have to type in order to choose the local one…

> Not sure what you mean by combining list-layouts and size? I don't see the point of complicating the help of list-layouts to allow you to select a certain size. You can grep for it. Or you meant something else?

Yes. It’s what I meant. I think, if we have `--list-layouts` it should default to the ones that can be used in a regular play. That is, it should default to the ones with normal size. This makes the most sense to me. If you want to list non-standard ones, you should combine it with `--size all`. It is a very natural combination of command line arguments to me.

otizonaizit:
>If you just copy it over, it will have a suffix `.layout` anyway that you will have to type in order to choose the local one…
>
>>Not sure what you mean by combining list-layouts and size? I don't see the point of complicating the help of list-layouts to allow you to select a certain size. You can grep for it. Or you meant something else?
>
>Yes. It’s what I meant. I think, if we have `--list-layouts` it should default to the ones that can be used in a regular play. That is, it should default to the ones with normal size. This makes the most sense to me. If you want to list non-standard ones, you should combine it with `--size all`. It is a very natural combination of command line arguments to me.

Good luck explaining that semantics in the help! Again, no one uses list-layouts, that's only there for reference.

Debilski:
I wouldn’t explain it. I didn’t even know it was still there.

Should we also make it so that `--layout 1` loads `normal_001` (according to `--size` of course)? Saves some typing.

otizonaizit:
>I wouldn’t explain it. I didn’t even know it was still there.
>
>Should we also make it so that `--layout 1` loads `normal_001` (according to `--size` of course)? Saves some typing.

Nope, that's also hard to explain. Come on, you specify one file or one name, and that's it, no need for short-cuts. No one is typing this names so often that you need to save typing. Except you, but you can of course link local files 1...100 to pelita/_layouts/normal001..100 and be done with it ;)

Debilski:
> Except you, but you can of course link local files 1...100 to pelita/_layouts/normal001..100 and be done with it ;)

Nice idea but it doesn’t work in combination with `--size` ;p

I am pretty sure, I have students seen painfully typing every layout after another when testing. (And the kind of students who don’t know that they could use the arrow-up key in bash.)

But you are right it’s harder to explain. I guess it should probably be the only way to specify a built-in layout. I mean, why `normal_001` if normal is the default anyway? (But we can postpone this.)

otizonaizit:
>>Except you, but you can of course link local files 1...100 to pelita/_layouts/normal001..100 and be done with it ;)
>
>Nice idea but it doesn’t work in combination with `--size` ;p
>
>I am pretty sure, I have students seen painfully typing every layout after another when testing. (And the kind of students who don’t know that they could use the arrow-up key in bash.)
>
>But you are right it’s harder to explain. I guess it should probably be the only way to specify a built-in layout. I mean, why `normal_001` if normal is the default anyway? (But we can postpone this.)
>

But you see that you already say it yourself why the "shortcut" wouldn't help? If they don't know how to use the arrow-up key in bash, do you think they will learn that what it shows as "normal__001" can be specified as "1"? The names now are much shorter than before: the pain will be much less than last year, even for those who type every command line by hand.

The only compromise I could make is that we ditch the small_ and big_ layouts completely. I.e. we only have them for our own internal tests, but we don't advertise them, and we keep the size option hidden and undocumented. Students will *not* need to use big_ nor small_, surely not when manually testing with the TK gui.

Debilski:
>  do you think they will learn that what it shows as "normal__001" can be specified as "1"?

It should be shown differently, of course. But I am all for postponing this issue.

PR #672, opened 2019-08-14
Cleanup docs

otizonaizit:
- Removed the now obsolete `groupN` directory from the docs, so that we can finally run `pytest` again without specifying any argument. 
- Update the screeshot of the game, so that it matches the newest UI.
- Changed players for the pelita CLI demo game: use the best builtins we have, so that the game is more interesting to watch


Debilski:
The SmartRandomPlayers are our best builtins?

Debilski:
Your other cleaning efforts are outdated but ok :)

Debilski:
But seriously. Why the SmartRandomPlayers?

otizonaizit:
>But seriously. Why the SmartRandomPlayers?
>
Change it to something else. I don't care.

PR #675, opened 2019-08-15
NF: Add shadow bots to Tk UI

### LINKED ISSUE #347: Show noisy ghost positions in UI
Have the UI optionally show the position of where the currently moving bot has experienced the other bots when their position is noisy.

Debilski:
We’ll need to spend some more time to make it look less intrusive.

Closes #347

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -367,6 +367,10 @@ def setup_game(team_specs, *, layout_dict, max_rounds=300, layout_name="", seed=
         # List of boolean flags weather bot has been eaten since its last move
         bot_was_killed = [False]*4,
 
+        # The noisy positions that the bot in `turn` has currently been shown.
+        # None, if not noisy
+        noisy_positions = [None] * 4,
+
         #: Messages the bots say. Keeps only the recent one at the respective bot’s index.
         say=[""] * 4,
 
@@ -465,6 +469,7 @@ def request_new_position(game_state):
 
     bot_state = prepare_bot_state(game_state)
 
+
     start_time = time.monotonic()
 
     new_position = move_fun.get_move(bot_state)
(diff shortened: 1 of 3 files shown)


otizonaizit:
Is this ready to be merged? If yes, I would like to use a screenshot in working on https://github.com/ASPP/pelita_template/issues/23 . If it is not mentioned in the pelita_template docs it shoudln't go in, in my opinion. It is a nice addition, so I would vote for merging, even if it is ot perfect now. You can open an issue where you explain what you would like to have from a better version.

otizonaizit:
OK, after playing around with it I am not for merging anymore. The shadow bots are showed unconditionally. They should be shown only when the grid is on. (And the grid button renamed to `debug`). Like it is now it is unsuable, as it makes the normal game quite chaotic to follow.
And, when running step-by-step you'll that the shadow bots are not always shown when they should.
Given that if it goes in it needs to be documented, either you can get it ready in the next days, or we just postpone it for after ASPP 2019.

Debilski:
> On 16. Aug 2019, at 10:43, Tiziano Zito <notifications@github.com> wrote:
> 
> Is this ready to be merged?
> 

Not yet.

Debilski:
Leave it out for now. If we get it ready, we can just update the images, no?

Debilski:
Current design: image

I am still planning to add the diamond of noise and probably some arrows to better show the movement that the current bot has taken, which hopefully doesn’t overcomplicate things.

otizonaizit:
These look very good, thanks! I think the "show-grid" button should be renamed to "debug" now that it displays so much more than just the grid. But I think this could be merged already as-is: https://xkcd.com/974/

Debilski:
Yeah, makes sense to rename it now. Although for the shadow I was thinking about adding an additional button. Guess we’ll have to play with the debug mode a little more to see if it maybe is a little too much information all at once.

otizonaizit:
I think there are already too many buttons. I think the best is to have the normal mode and the debug mode, and in debug mode you see everything there is too see. If the noise diamonds are equally unintrusive like you did the noisy bots I think it's not going too crowded.

PR #674, opened 2019-08-15
document user-facing functions in utils

otizonaizit:
Document user-facing functions

Fixes: #671

PR #676, opened 2019-08-15
Run games in headless mode and replay them later

Debilski:
Users can play a game in headless mode, store its result and additionally the history of game states during the game, which is useful, if they want to know what went wrong during a game.

    from pelita.utils import run_background_game, replay_background_game

    result, replay = run_background_game('0', '1')
    
    replay_background_game(replay)
    # opens Tk viewer

    replay_background_game(replay, viewer='ascii')

Still needs a bunch of tests and utils suffers from circular imports.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -621,19 +621,18 @@ def apply_move(gamestate, bot_position):
     """
     # TODO is a timeout counted as an error?
     # define local variables
-    bots = gamestate["bots"]
+    bots = gamestate["bots"][:]
     turn = gamestate["turn"]
     team = turn % 2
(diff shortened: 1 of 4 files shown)


otizonaizit:
I think this is overly complicating things. And, especially if they use the background games to gather statistics, which is was most people asked for, keeping all intermediate states will soon use a lot of memory. I addressed replaying a background game in https://github.com/ASPP/pelita_template/pull/27/commits/1e7ec326b8c889bc04a3744ddf2911b561c17df0 , which requires an explicit handling of the seed in `run_background_game` which is done in https://github.com/ASPP/pelita/pull/674/commits/6172522949862fbae7510824b92817fd77a4185d

PR #682, opened 2019-08-16
Layout refactoring

otizonaizit:
Get rid of the `create_layout` function and unify layout parsing in `layout.py`. 

Now that the code is all in one place it could be improved and streamlined. To avoid disruption right before ASPP-2019, I propose to postpone the refactoring to after ASPP-2019.


otizonaizit:
Note that this PR is to be merged after #674 .

otizonaizit:
Note that the travis failure is expected. That test is testing something I don't think we want to enforce. If the PR goes in I will remove the failing test

PR #687, opened 2019-09-02
ENH: Add flag to silence run_game

### LINKED ISSUE #679: setup_viewers shouldn't add ResultPrinter unconditionally to the list of viewers
Here https://github.com/ASPP/pelita/blob/6146c1db6142cedecf4ed0be21069d59a7036115/pelita/game.py#L159
The ResultPrinter viewer is added unconditionally to the list of viewers. It shouldn't. If the list of viewers is empty, then nothing should be printed. The problem appears when trying to run background games with `run_background_game`: the final result is printed. If running a lot of games this is annoying.
Not sure if this can be simply added only when `viewers is not None`, because it is not clear to me if the tournament code needs this.

Debilski:
Closes #679

Need test (possible rename) and will need to be added to the scripts in utils.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -83,7 +83,8 @@ def controller_exit(state, await_action='play_step'):
 
 def run_game(team_specs, *, layout_dict, layout_name="", max_rounds=300, seed=None,
              max_team_errors=5, timeout_length=3, viewers=None, viewer_options=None,
-             store_output=False, team_names=(None, None), allow_exceptions=False):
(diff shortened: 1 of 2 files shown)


Debilski:
Q1: Maybe silent mode should be the default?
Q2: Should this also silence the `ᗧ blue team '0' -> 'Smart Eating Players'` line from pelita-player?

Debilski:
@otizonaizit your thoughts on adding this to v2.0?

PR #690, opened 2019-09-07
Do not return state from move()

keszybz:
Things are simpler if we always initialize the state to be an empty
dictionary, and simply pass the same mutable object always to the move
function. In cases where the player does not actually use the state to
store anything, it can ignore the variable completely. The object also
has always the same type (dict), so there is less chance of trying to
a TypeError.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -92,10 +92,10 @@ def run_game(team_specs, *, layout_dict, layout_name="", max_rounds=300, seed=No
     team_specs : list[team0, team1]
               a list specifying the two teams to play the match with. The list
               items team0 and team1 must be either
-               - a function with signature move(bot, state) -> (x, y), state
-                 the function takes a bot object and a state and returns the next
-                 position (x, y) of bot and state.
-                 state can be an arbitrary Python object, None by default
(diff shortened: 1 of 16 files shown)


Debilski:
I’ll think about it in the coming days. Maybe it’s not as bad as I thought but I am still unsure about if it is nice enough to use with testing.

otizonaizit:
Thanks @keszybz , after talking with you and having seen the students struggling with it, I think it is a good idea to fix the state object to be a dictionary, and not to return it in the move function. I am +1 on this!

otizonaizit:
Before merging, please also address: https://github.com/ASPP/pelita_template/pull/33

keszybz:
Updated to remove the last cases of `if not state`. It is now `if state == {}` everywhere.

keszybz:
Anything else to do here?

otizonaizit:
LGTM

keszybz:
Anything blocking this?

Debilski:
Uhm, multiline `say` … do we want to allow this?

keszybz:
It works, mostly. The font I got was not monospace, so the "underlining" was a bit off.  Maybe some different example would be better.

PR #697, opened 2019-10-11
BLD: Bump version to 2.1-pre to signalise recent API changes

Debilski:
Not sure if it’s really needed but it could avoid some confusion. A proper release should include a few more fixes than just the API change.


Debilski:
Merging this. Additionally, I tagged and pushed a v2.0.1 bugfix version to pypi (that does not include the API changes) which fixes the geometry bug.

PR #698, opened 2019-11-13
ENH: Use runpy for importing a player.

Debilski:
*draft PR*

I stumbled across this tweet https://twitter.com/GaelVaroquaux/status/1194336723469316096 which suggests the [`runpy`](https://docs.python.org/3/library/runpy.html) module for importing external code (also see: https://stackoverflow.com/questions/436198/what-is-an-alternative-to-execfile-in-python-3/24261031#24261031), which void avoid a few of the irregularities our `importlib` solution currently has.

However, on the top of the module it says

> Furthermore, any functions and classes defined by the executed code are not guaranteed to work correctly after a runpy function has returned. If that limitation is not acceptable for a given use case, importlib is likely to be a more suitable choice than this module.

and I wonder how relevant this is for us and what this implies. Any thoughts?

Description of the commit:

Previously, we used our own hack for importing a user’s Python file for a
player, which involved modifying sys.path to include the directory of
the Python file and then import the Python file as a module with
importlib.import_module. This came with some mild (and sometimes
unexpected) annoyances, namely, it was not possible to import a file
from a module with the same name of a previously imported module.

A player in a directory called ‘sys’, or – more likely – ‘pelita’ would
therefore never be found by the program. It was also not possible to
import two players from different but equally-named folders, although
this problem had been circumvented by running each player in a single
subprocess always.

This commit changes the import mechanism to using the runpy module for
importing the player’s module, TEAM_NAME and move function.

### DIFF
--- a/pelita/scripts/pelita_player.py
+++ b/pelita/scripts/pelita_player.py
@@ -2,10 +2,10 @@
 
 import argparse
 import contextlib
-import importlib
 import logging
 import os
 from pathlib import Path
+import runpy
 import signal
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


Debilski:
Closing for now because of the caveat mentioned above.

PR #702, opened 2020-02-14
Merge pelita_template repo

### LINKED ISSUE #692: Merge pelita_template back in to the core
Now that the API rewrite is ready, I'd like to move the demos in https://github.com/ASPP/pelita_template back to the core. This makes cross-references and testing easier. The https://github.com/ASPP/pelita_template/blob/master/README.md file should become the core of the new documentation.

Note that this is not a problem for the group repos: we initialize them anyway by just copying the content of `pelita_template`. We can as well copy the contents of `pelita/demos`.

@keszybz @Debilski : what do you think?

Debilski:
Closes #692.

This merges the whole of the https://github.com/ASPP/pelita_template repo into a subfolder `template/`.
To avoid the complete mess of merging the template repo directly and then moving it into the subfolder, I have taken the liberty of filtering the repo first with

    git filter-repo --to-subdirectory-filter template

and then merging with `--allow-unrelated-histories`.

This means that tools such as `git blame` will work but of course, once one checks out a commit from the template branch, the commit will *only* contain the `template/` folder, which may be somewhat weird. Alternatively, we could *insert* the `template/` folder’s history on top of the current master, although of course, older commits from the template will not work together with the current pelita, so this is also probably a bad idea.

Before I merge, I would like to know, if this is a good idea (and if the folder’s name should be  `template` or something else — let’s fix that now while we’re rebasing anyway.)

TODO: Tests should be activated again once we have decided if this is the right approach.


Debilski:
@otizonaizit I need a second opinion if this is a viable approach and will then resubmit, after we’ve merged the layout unification.

otizonaizit:
hmm. 

I prefer the option where the commits are inserted on top of master (if possible first on a branch, so that we have a nice loop in the history and can get rid of the whole thing by removing the merge commit). It is true that there is no guarantee that checking out an old commit will allow you to run the code in template, but at least you have a chance.

The "merging" of repo history in the approach you implemented here doesn't seem to me it is giving us any advantage. In the sense that if you step back in history you won't have the pelita stuff in the working directory, so you can't run it for sure. The commits will be following a more realistic timeline, but I don't think it is so important to have that.

Debilski:
> On 6. Jun 2020, at 12:17, Tiziano Zito <notifications@github.com> wrote:
> 
> ﻿
> hmm.
> 
> I prefer the option where the commits are inserted on top of master (if possible first on a branch, so that we have a nice loop in the history and can get rid of the whole thing by removing the merge commit). It is true that there is no guarantee that checking out an old commit will allow you to run the code in template, but at least you have a chance.
> 

The chance is 99% that it won’t work though. :) But you have a point in that it makes things a little easier to understand when it would run with a 1 line fix (like in 20% of the cases); for all other cases one will have to use two repos anyway.

> The "merging" of repo history in the approach you implemented here doesn't seem to me it is giving us any advantage. In the sense that if you step back in history you won't have the pelita stuff in the working directory, so you can't run it for sure. The commits will be following a more realistic timeline, but I don't think it is so important to have that.
> 

I’m fine with either way. I put this approach in the PR more as a proof of concept, because it was the slightly harder approach. I’ll resubmit the other one for show later or early next week.

PR #703, opened 2020-02-15
BF: Disable randomized hash seeds to avoid introducing non-predictability.

### LINKED ISSUE #689: hash seed needs to be fixed or reproducibility is not guaranteed
Or something like that. I am not sure but @otizonaizit knows what is going on.

Debilski:
Randomized hash seeds (the default in recent Python versions) make the
game non-reproducible, as different Python runs will have a different
results for code such as list(set(x)), which is independent of the
manually seeded rng.

We now disable randomized seeds completely by setting an environment
variable PYTHONHASHSEED='0' before calling the Pelita players. (A future
version could manually set different seeds for the individual players,
dependent on the global seed value.)

Caveat: This only solves the problem for _external_ players that are
started by the main Pelita executable. Internal players will have a
different set ordering each time the main program is run.

Closes #689.

### DIFF
--- a/pelita/player/team.py
+++ b/pelita/player/team.py
@@ -250,10 +250,12 @@ class RemoteTeam:
             # We must run in unbuffered mode to enforce flushing of stdout/stderr,
             # otherwise we may lose some of what is printed
             proc = subprocess.Popen(external_call, stdout=stdout, stderr=stderr,
-                                    env=dict(os.environ, PYTHONUNBUFFERED='x'))
+                                    env=dict(os.environ,
+                                             PYTHONUNBUFFERED='x',
+                                             PYTHONHASHSEED='0'))
             return (proc, stdout, stderr)
         else:
-            return (subprocess.Popen(external_call), None, None)
+            return (subprocess.Popen(external_call, env=dict(os.environ, PYTHONHASHSEED='0')), None, None)
 
     @property
     def team_name(self):
--- a/test/test_remote_game.py
+++ b/test/test_remote_game.py
@@ -246,3 +246,70 @@ def test_remote_dumps_with_failure(failing_team):
 
     # No errors for red
     assert good_err == [""]
+
+
+def test_remote_reproducible():
+    # Test that the rng gives reproducible results,
+    # even when sets are involved.
+    #
+    # Currently, we have disabled hashed seeds completely (PYTHONHASHSEED='0').
+    # If we make this dependent on the game_state’s internal seed,
+    # we should run this test several times with different starting seeds.
+    layout = """
+        ##############
+        #            #
+        #   2    3   #
+        #            #
+        #            #
+        #0    ..    1#
+        #            #
+        ##############
+        """
+
+    player = dedent("""
+    def random_player(bot, state):
+        # somehow list(set(bot.legal_positions)) seems to be quite
+        # stable even when randomized hashes are activated
+        # however, list(set(["A", "B", "C", "D"])) very consistently
+        # gives random results. We try to mix both tests to
+        # increase our chances of detecting failure
+
+        lettertest = list(set(list("ABCD")))
+
+        poss = list(set(bot.legal_positions))
+        if lettertest[0] == "A":
+            return bot.position
+        else:
+            return poss[0]
+
+    move = random_player
+    TEAM_NAME = "random test team"
+    """)
+    out_folder = tempfile.TemporaryDirectory()
+
+    testseed = random.randint(0, 20000)
+    num_tries = 6
+    all_paths = []
+
+    for num in range(num_tries):
+        # store all positions that the random bots are taking
+        paths = {i: [] for i in range(4)}
+
+        with tempfile.NamedTemporaryFile('w+', suffix='.py') as f:
+            with tempfile.NamedTemporaryFile('w+', suffix='.py') as g:
+                print(player, file=f, flush=True)
+                print(player, file=g, flush=True)
+
+                state = pelita.game.setup_game([f.name, g.name],
+                                               max_rounds=20,
+                                               layout_dict=pelita.layout.parse_layout(layout),
+                                               seed=testseed,
+                                               allow_exceptions=True)
+
+                while not state['gameover']:
+                    # play the next turn
+                    state = pelita.game.play_turn(state, allow_exceptions=True)
+                    # add current bot position to paths
+                    paths[state['turn']].append(state['bots'][state['turn']])
+        all_paths.append(paths)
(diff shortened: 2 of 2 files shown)


Debilski:
Some update on this issue.

1) Recent versions of Python (CPython; de facto from 3.6, de jure from 3.7) keep the insertion order of a `dict` when iterating. (https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects)
2) This guarantee does not hold for the standard `set`.
3) Iteration through a set seems to be defined by the hash value of its objects.
4) Hash values of (reasonably small) `int`s are deterministic. (`hash(i) = I`)
5) Hash values of strings are salted with the `PYTHONHASHSEED`.
6) Hash values of objects may (or may not? – see 8.) be salted with the `PYTHONHASHSEED`.
7) Hash values of tuples are deterministic and depend only on the hashes of their members.
8) The default hash values of an `object` depends on its `id()`, which is arbitrary in each new process.
9) It therefore follows that setting `PYTHONHASHSEED=0` only has an effect when applied to sets of strings (and some other types) but never on sets of arbitrary objects or ints or tuples of ints (coordinates).

The graph implementation from MDP, that the BayesAvenger bots require, uses `GraphNode` objects that come without a `__hash__` (and `__eq__`) method. (https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/graph/graph.py#L46) Non-predictability is introduced, when the list of nodes is at one point converted to a set.
One of two changes may fix the problem. a) Add the missing methods to MDP, or b) use a self-implemented [`OrderedSet`](https://stackoverflow.com/a/1653978) or a dict in the bot’s code.

Given that `PYTHONHASHSEED=0` is only effective on sets of strings, we’ll have to decide if this is a big enough issue to apply this PR.
As this change won’t make `run_background_game` reproducible (again the only proper fix is to use some kind of OrderedSet or dict), I guess we may as well close this issue.

otizonaizit:
I wouldn't close the issue still. We need to document this, and also make sure we don't have anything within pelita which depends on the order of sets. I think reproducibility of games is an essential feature that we have to strive for.
We can't control external players, but we have to document this for users, so that they know how to avoid this issue

otizonaizit:
> The graph implementation from MDP, that the BayesAvenger bots require, uses `GraphNode` objects that come without a `__hash__` (and `__eq__`) method. (https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/graph/graph.py#L46) Non-predictability is introduced, when the list of nodes is at one point converted to a set.
> One of two changes may fix the problem. a) Add the missing methods to MDP, or b) use a self-implemented [`OrderedSet`](https://stackoverflow.com/a/1653978) or a dict in the bot’s code.

this is very good detective work, thanks!!! We spent hours and hours trying to understand this issue!

otizonaizit:
> The graph implementation from MDP, that the BayesAvenger bots require, uses `GraphNode` objects that come without a `__hash__` (and `__eq__`) method. (https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/graph/graph.py#L46) Non-predictability is introduced, when the list of nodes is at one point converted to a set.
> One of two changes may fix the problem. a) Add the missing methods to MDP, or b) use a self-implemented [`OrderedSet`](https://stackoverflow.com/a/1653978) or a dict in the bot’s code.
>

I think the best fix would be for BayesAvenger to use networkx instead of MDP. And make sure then that with networkx everything is reproducible. This may at some point graduate BayesAvenger as show-off bots for Pelita.

Debilski:
> I wouldn't close the issue still. We need to document this, and also make sure we don't have anything within pelita which depends on the order of sets. I think reproducibility of games is an essential feature that we have to strive for.
> We can't control external players, but we have to document this for users, so that they know how to avoid this issue

Ok, I’ll document the problem (and maybe also add some updates to the BayesAvengers).

Pelita itself should be fine even with #708 merged. All our sets in the main game only contain `Tuple[int]`.

otizonaizit:
To future memory, here a little script that shows the problem of non-reproducibility of set ordering:

```python
import multiprocessing

class A:
    def __init__(self, i):
        self.i = i
    def __repr__(self):
        return str(self.i)

def f(n):
  print(set([A(i) for i in range(n)]))

if __name__ == '__main__':
    with multiprocessing.Pool(2) as p:
        p.map(f, [10, 10])
```

On my machine right now this prints:

```
{5, 0, 6, 1, 7, 2, 8, 3, 9, 4}
{8, 3, 9, 4, 5, 0, 6, 1, 7, 2}
```
It is necessary to use multiple processes, because re-running the same code in the same terminal typically re-uses the memory allocations and so it appears that the set ordering is preserved.
Even with this script, if you run it multiple times, you may get the very same order of the two sets. After reboot you'll get a different order.

lschwetlick:
Just confirming that on my machine too, your code gives me two different sets!

Debilski:
Closing this PR without merging. I think we now understand the tradeoffs a bit better and at least know *when* it may happen. Setting the `PYTHONHASHSEED` manually would only introduce another moving variable here (as it would only work for network games but not, eg. when trying to replicate a failed `run_background_game` attempt).

otizonaizit:
A [recent discussion](https://mail.python.org/archives/list/python-dev@python.org/message/73KMUPBTS4MIOMPRT3PBQ36HREQFXUUN/) on  [python-dev](https://mail.python.org/mailman3/lists/python-dev.python.org/) seems to indicate that `hash(None)` may become deterministic in the near future, which implies that set order will become deterministic, which means that we may revise this discussion.
I still haven't given up completely on replicable pelita games ;-)
Commenting on this pull request so that we do not forget to come back here if the Python developers come to an agreement (I hope they do).

Debilski:
Not sure I understand all the reasonings there (also I don’t want to read it all), but just fixing `None` won’t save us, no?
We need sets with insertion order and this will fix at least this very specific problem of iteration.
Code that uses `hash(object())` and therefore `id(object())` in any other comparison operation will by design be unfixably indeterministic (unless it overrides `hash`), so I am somewhat less confident than you that we will ever reach your goal. :)

otizonaizit:
yes, we won't have 100% replicability ever, but set order is one of the last roadblocks with built-in data structures. We can't save people who write custom classes with funky hashes, but using sets is pretty legitimate and should be even encouraged in pelita for all things where the natural data structure is a set. So, if they fix set order I think we should enable PYTHONHASHSEED="some seed" and we are good to go.

Debilski:
Yeah, insertion order for sets will be good enough in many cases. Not sure if we should need `PYTHONHASHSEED` anyway, as it introduces other edge cases (only works in network games). But this is a discussion to be had in the future. :)

PR #708, opened 2020-06-25
Adding shape to layout dictionary and optimising access

### LINKED ISSUE #673: Add width and height to the layout dictionary
As it is needed in several places, it makes sense to just stick two additional keys into the layout dictionary. Fixing this bug also means getting rid of all `max(walls)[0]+1`, `max(walls)[1]+1`, `sorted(walls)[-1][0]`, `sorted(walls)[-1][1]` spread in the pelita and pelita_template code.

Debilski:
Work in progress (discussion) PR. **Do not merge yet!** Closes #673.

### Current problems

This PR tries to optimise two critical bottlenecks in the Pelita backend:

  * It is costly and error prone to get the shape of the maze, which now uses the pattern `width = max(walls)[0] + 1`, `height = max(walls)[1] + 1`. And this has to be repeated every time we use it. The `+1` is easily forgotten. As we are sorting the walls, it might be enough to assume `max(walls) == walls[-1]` to speed this up.
  * It is costly to check whether a position is in a wall (or in the food list) (eg. the pattern `if position in gamestate["walls"]`), and even costlier to check when it’s not, as the whole list needs to be traversed every time.

Both problems can be solved by adding a tuple `shape` to the layout and by using a `(frozen)set` instead of a `list` for the walls (and maybe also food).
Tricky parts involve changing some of our helper functions that only rely on the `walls` to do their job (eg. the noiser), which now need to receive an additional argument `shape`.

### Outline of this PR

This draft PR proposes the following outline to get there:

- [x] Have a benchmark in `demo/benchmark_game.py` that uses `timeit` and `cProfile` to ensure the speed gains are worthwhile. (Changes to the benchmark should be made in their own commit and rebased to an earlier position in the history, so that the benchmark code stays the same throughout the changes of Pelita.)
- [x] Add `shape` to the output of `parse_layout` and include it in the `gamestate`.
- [x] Replace usages of the `max(walls) + 1` pattern with shape
- [x] Add/update tests
- [x] Remove code duplication in `team.py` (walls/homezones are created multiple times per turn for each of the virtual bots in `make_bots`). Many of those can be cached in `class Team`.
- [x] Make `walls` a `frozenset`
- [x] Make `homezone` a `frozenset` or make it a function that just uses the `shape` for the `if pos in homezone` pattern
- [x] Make `food` a `frozenset` (if it makes sense)

### Drawbacks of walls being a `frozenset`

Patterns that are iterating the maze as a whole can be optimised more easily with a (sorted) list. It is assumed that these are only relevant for printing and therefore negligible.

### Alternatives

The only other alternatives would be to use a numpy object or a specialised class (as we had in the old pre-2.0 datamodel) that would be optimised for rectangular access. In the spirit of using only built-in types in the gamestate, those alternatives have to be rejected.

### Benchmarking

The benchmark includes 4 different game constellations, local and remote. The following code will run 5 games directly after another and repeat this 3 times. The fastest result is returned.

    python demo/benchmark_game.py --repeat 3 --number 5

To see in which functions we spend most of the time, cProfile is helpful. Use it with a number to specify the team constellation.

    python demo/benchmark_game.py --cProfile 0 # or 1, 2, 3

### Open question

Is it a good idea to split the walls and the shape? Or should we have a dict/dataclass for the layout: `{"walls": [...], "shape": (16, 32)`?

### DIFF
--- a/check_layout_consistency.py
+++ b/check_layout_consistency.py
@@ -1,35 +1,10 @@
 """Detect if a layout contains "chambers" with food"""
 import sys
 
-import networkx
+import networkx as nx
 from pelita.layout import parse_layout
+from pelita.utils import walls_to_graph
 
-
-def get_hw(layout):
-    walls = layout['walls']
-    width = max([coord[0] for coord in walls]) + 1
-    height = max([coord[1] for coord in walls]) + 1
-    return height, width
-
-def layout_to_graph(layout):
-    """Return a networkx.Graph object given the layout
-    """
-    graph = networkx.Graph()
-    height, width = get_hw(layout)
-    walls = layout['walls']
-    for x in range(width):
-        for y in range(height):
-            if (x, y) not in walls:
-                # this is a free position, get its neighbors
-                for delta_x, delta_y in ((1,0), (-1,0), (0,1), (0,-1)):
-                    neighbor = (x + delta_x, y + delta_y)
-                    # we don't need to check for getting neighbors out of the maze
-                    # because our mazes are all surrounded by walls, i.e. our
-                    # deltas will not put us out of the maze
-                    if neighbor not in walls:
-                        # this is a genuine neighbor, add an edge in the graph
-                        graph.add_edge((x, y), neighbor)
-    return graph
 
 if __name__ == '__main__':
 
@@ -44,8 +19,8 @@ if __name__ == '__main__':
 
     # first of all, check for symmetry
     # if something is found on the left, it should be found center-mirrored on the right
-    height, width = get_hw(layout)
-    known = layout['walls']+layout['food']
+    width, height = layout['shape']
+    known = layout['walls'] | set(layout['food'])
     layout['empty'] = [(x,y) for x in range(width) for y in range(height) if (x,y) not in known]
     for x in range(width // 2):
         for y in range(height):
@@ -56,14 +31,14 @@ if __name__ == '__main__':
                         print(f'{flname}: Layout is not symmetric {coord} != {cmirror}')
 
 
-    graph = layout_to_graph(layout)
+    graph = walls_to_graph(layout['walls'])
 
     # check for dead_ends
     for node, degree in graph.degree():
         if degree < 2:
             print(f'{flname}: found dead end in {node}')
 
-    if networkx.node_connectivity(graph) < 2:
-        entrance = networkx.minimum_node_cut(graph).pop()
+    if nx.node_connectivity(graph) < 2:
+        entrance = nx.minimum_node_cut(graph).pop()
         print(f'{flname}: Detected chamber, entrance: {entrance}')
 
--- a/demo/benchmark_game.py
+++ b/demo/benchmark_game.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import argparse
+import cProfile
 import functools
 import timeit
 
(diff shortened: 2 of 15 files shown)


Debilski:
Here’s an in-progress update on the speed:

First with the background_games demo.

```
./pelita_template (master)
❯ time python demo08_background_games.py
Games played: 100
Attacker wins: 17
Defender wins: 79
Draws: 4
Average score difference: -14.71

________________________________________________________
Executed in   92.99 secs    fish           external
   usr time   90.31 secs  203.00 micros   90.31 secs
   sys time    0.64 secs  898.00 micros    0.63 secs
```

```
./pelita_template (this PR)
❯ time python demo08_background_games.py
Games played: 100
Attacker wins: 18
Defender wins: 77
Draws: 5
Average score difference: -18.29

________________________________________________________
Executed in   20.08 secs    fish           external
   usr time   19.55 secs  276.00 micros   19.55 secs
   sys time    0.32 secs  959.00 micros    0.32 secs
```


Now with the benchmark that I created for this PR:

```
❯ python demo/benchmark_game.py --number 5 --max-rounds 300
Running 5 times with max 300 rounds. Fastest out of 5:
Stopping            : 0.6716697419999997
NQ_Random           : 0.9012290820000004
Stopping (remote)   : 4.812962928999998
NQ_Random (remote)  : 5.1260686190000015
```

```
❯ python demo/benchmark_game.py --number 5 --max-rounds 300
Running 5 times with max 300 rounds. Fastest out of 5:
Stopping            : 4.707969789
NQ_Random           : 6.251112241999998
Stopping (remote)   : 10.743467068000001
NQ_Random (remote)  : 11.85318602800001
```

As we can see, the new PR can make the code between 2 times and up to 8 times faster in the best case in a local game. Of course, the optimisations are only relevant for code that accesses the shape (this won’t happen too often to be a relevant factor) and checks `bot.walls`. As many good bots will at some point switch over to using some sort of a graph, they also shouldn’t need to access `bot.walls` more than once (and `bot.legal_positions` wouldn’t be needed either), but our own code already profits a lot from the changes.


My suggestion would be to merge this in the coming days (I still want to have another look at the code and docs) with the current changes:

  * added a `shape` to `game_state` and `bot` to avoid having to use the costly and error-prone `max(walls)[0]+1`
  * made `walls` a `Set[Tuple[int]]`
  * made `homezone` a `Set[Tuple[int]]` (we can still make this a special object with `__contains__` syntax later)

Notably, `walls` and `homezone` are no `frozenset`s yet. Also, given the problems with `random.choice`, `food` is still a `List[Tuple[int]]`. I haven’t checked it, but the access pattern for food might not profit from it being a set *that* much. (Arguably, the *usual* access pattern for food also doesn’t include `random.choice`, but I think it’s still useful for experimenting.)


I have rewritten Pietro’s bots to use networkx while using this branch and haven’t really run into problems (one usual change was the need to do `homezone[:]` → `homezone.copy()` but that’s a minor thing). I’d say it is safe to use.

Debilski:
Merging this now. Whether we use `frozenset`s is still open for discussion. Also, please note that in the main `game_state` we already have the food stored in (two) sets. This is however converted to a list when it gets passed to the teams.

There are still optimisations possible in how the `Bot` objects are created. (Caching and code cleanup.) And eventually, we could add an lru cache (#719) to the noiser, as estimating the free positions at each step is one of the most costly things in the backend.

PR #709, opened 2020-08-21
Put max_team_errors in game state

Debilski:
Previously, this was only a global variable `MAX_ALLOWED_ERRORS ` in `game.py`, even though we did have an empty, non-used API slot (with a default) in `def run_game` and `def setup_game` and an argument in the command line interface `--max-timeouts` that was not functional either.

I have rewired everything to make use of the flag and renamed the CLI to `--max-errors`, which is hopefully more clear. Note that I also changed the defaults to 4, as in my opinion the name `--max-errors` implies the maximum number of errors that are still allowed and our rule is ‘(minimum) five timeouts and you’re out’ (open for suggestions to change the wording and default to be 5 again, but `--num-errors-to-die` does not sound much nicer).

For running games without error checking (might be useful for testing), I’ve added a branch that disables the error check when a negative number is passes. We don’t need to have this but I find it a little nicer than having to invent arbitrarily big numbers (and probably not important enough to include a flag `--no-max-errors` – or would this be better?).

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -901,8 +901,11 @@ def check_gameover(game_state, detect_final_move=False):
 
     # If any team has reached more than max_team_errors errors, this team loses.
     # If both teams have reached more than max_team_errors errors, it’s a draw.
+    # If max_team_errors is < 0, the game will go on without checking.
     num_errors = [len(f) for f in game_state['errors']]
-    if num_errors[0] <= game_state['max_team_errors'] and num_errors[1] <= game_state['max_team_errors']:
+    if game_state['max_team_errors'] < 0:
+        pass
+    elif num_errors[0] <= game_state['max_team_errors'] and num_errors[1] <= game_state['max_team_errors']:
         # no one has exceeded the max number of errors
         pass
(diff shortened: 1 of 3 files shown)


otizonaizit:
Can we make the number for disabling error checking to be "0" instead of a negative number? That's a typical flag-number and it is also easier to document and to remember:

--max-errors N  max number of errors before dying. Defaults to 4. Set to 0 to disable error checking altogether

Something like that.

Debilski:
Ok, but then I’d suggest using a different name for the variable without ‘max’ in it. `--error-limit`?

otizonaizit:
Sure, why not, good name. Maybe even simpler would be just "errors"? The text documentation is needed anyway to understand what the parameter does, it's not like we can encode in the name alone the function.

Debilski:
`errors` would also be fine to use on the command line, but we already use in the name in the game state for collecting the team errors. I’ll add a PR that changes it to `error_limit`/`--error-limit` for now. Then we at least have the number logic fixed. We can then discuss the definite naming at a later point.

otizonaizit:
OK, but also please use "0" to mean "no error checking".

Debilski:
Yes, of course.

PR #715, opened 2021-06-24
Add line of sight and intention arrows

Debilski:
(Work in progress. Will be rebased and become more refined eventually.)

This adds a bot’s line of sight in debug mode. Any enemy bot within the line of sight will appear at its exact position (and therefore will not have a ghost bot appear).

Note that the line of sight is drawn from a bot’s position *before* a move, while the UI will show it’s position *after* a move. To make this more clear, we show a border around the initial cell (or maybe we should colour the cell; will still experiment with this) and put an arrow in the direction of the move that was requested.

image

When a move was invalid and a random move is executed instead, we show an X on the original cell (design need to be refined).

image

How should we show that a move was a stopping move? Perhaps with this circular arrow? (This version comes with an X as it was also a random move because of an error.)

image

Even better would be, if we could animate the bots in Tk. This would, however, require a medium-sized rewrite of the Tk code.

**Technical details**

This is implemented with a new informational attribute in the game state `'requested_moves'` (needs tests), which informs the UI of the previous position and the position that has been requested (which can be different from the final position, as a bot can of course be eaten when moving – this might need a different kind of array, too).

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -377,6 +377,9 @@ def setup_game(team_specs, *, layout_dict, max_rounds=300, layout_name="", seed=
         # None, if not noisy
         noisy_positions = [None] * 4,
 
+        #: The moves that the bots returned. Keeps only the recent one at the respective bot’s index.
+        requested_moves=[None] * 4,
+
         #: Messages the bots say. Keeps only the recent one at the respective bot’s index.
         say=[""] * 4,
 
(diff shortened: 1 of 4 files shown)


Debilski:
If anyone wants to have a look before merging, I think it is good enough to go. UI can certainly be nicer but it’s the debug view after all …

Debilski:
(I hope I have the radius correct, this is more confusing than it should be …)

PR #738, opened 2021-10-05
BLD: Add Python 3.10 to CI

Debilski:


Debilski:
The improved error messages break a few of our test assumptions. Will fix it soon and append to this PR.

PR #739, opened 2021-10-11
BLD: Rearrange modules

### LINKED ISSUE #678: utils.py suffers from circular imports
The problem is that some built-in players are using the `walls_to_graph` function which is also a user-facing function. So the built-in players try to import utils right on `import pelita`.
On the other hand, the function `run_background_game` in `utils.py` uses `run_game`, which comes from `game.py`. So when you do `import pelita` you import the built-in players, which in turn need utils, which in turn need the `import pelita` to be finished already.
One solution would be not to import the built-in players on `import pelita`, but only when needed. The built-in players are used only in the demo game and in the tests, so importing them explicitly there seems a good solution.

Debilski:
Still WIP/RFC. Closes #678

This PR removes the explicit and implicit (via `pelita.players`) imports of `pelita.utils` in `pelita/__init__.py`, thereby resolving a circular import situation in `utils.py`.
The file `pelita/players/team.py` (which contains the `Bot` class as well as classes for the team state) was moved up in the hierarchy to `pelita/team.py` where it makes more sense to have it (apart from perhaps the `Bot` parts but that doesn’t matter for now).

As a bigger change, this means that a plain `import pelita` won’t be enough anymore for accessing `pelita.utils` (and for `pelita.players` but that is more relevant for internal use only).
Is it ok that we enforce the usage `import pelita.utils` or should we add some magic code to make that automatic but lazy?

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -1,20 +1,18 @@
 """This is the game module. Written in 2019 in Born by Carlos and Lisa."""
 
-import itertools
 import logging
 import os
 from random import Random
 import subprocess
 import sys
 import time
-import typing
 from warnings import warn
 
 from . import layout
 from .exceptions import FatalException, NonFatalException, NoFoodWarning, PlayerTimeout
 from .gamestate_filters import noiser
 from .layout import initial_positions, get_legal_positions
-from .network import bind_socket, setup_controller, ZMQPublisher
+from .network import setup_controller, ZMQPublisher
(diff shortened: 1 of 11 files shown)


otizonaizit:
I think this is long overdue, as the circular import thing was a mess, thanks! I also think that forcing `import pelita.utils` is not a problem at all: it is just a matter of changing the examples and users will simply follow.

PR #743, opened 2022-05-13
WIP: Adding support for zeroconf

Debilski:
This commit adds basic support for zeroconf-based networking.

Instead of a player location, pelita now understands the keyword `SCAN`
to search for remote players that advertise themselves via zeroconf:

    pelita ./own_player SCAN

A zeroconf-based remote player can be started with the argument
`--advertise` like so:

    pelita-player --advertise --remote ./some_bot.py tcp://192.168.1.23:10027

Demo:

https://user-images.githubusercontent.com/216179/168391205-0d130500-2295-47e0-92bb-9b4883e43d95.mov

Commit needs https://github.com/jstasiak/python-zeroconf and https://github.com/Textualize/rich (eye-candy) to run.

Not sure if we can call our service `_pelita-player._tcp`. For IANA correctness, we might want to create a uuid.

### DIFF
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -41,7 +41,7 @@ def scan(team_spec):
             if info:
                 addresses = ["%s:%d" % (addr, info.port) for addr in info.parsed_scoped_addresses()]
 
-                addr = f"remote:tcp://{addresses[0]}"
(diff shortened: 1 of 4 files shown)


Debilski:
Merging as a beta feature.

PR #747, opened 2022-09-08
BLD: v.2.2.0-rc1

Debilski:
- Do not have a team play two matches in a row during round-robin mode
 - Show the tournament group names in the UI and on the CLI
 - Automatic detection of network players via zeroconf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Pelita changelog
 
+  * v2.2.0 (???)
+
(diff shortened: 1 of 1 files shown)


Debilski:
Will create v2.2.0 if approved.

otizonaizit (approved):

otizonaizit:
The usage of zeroconf needs to be updated in the instructions [here](https://aspp.school/wiki/pelita) (in te commented out section, that will be enabled later in the week

PR #756, opened 2023-05-16
[WIP] New Qt UI (pyqt6)

Debilski:
Adding another work-in-progress PR for a different UI toolkit. This time PyQt6. Please see below for graphics.

I am undecided if we should really push this one and completely replace our old Tk design, but I have at least brought it into a workable state now. (I think the ghosts are still a little short and need a bit of a stretch.)

The advantages of Qt would be:

* Can do threading: We can receive zmq messages in a background thread and notify the app
* Much easier to do graphics: automatically scales all elements to the size that is needed
* Anti-aliasing
* Can easily export to PNG
* Does have an animation framework but I haven’t tested it yet.
* Might be easier to implement new features

Similarly to Tk:

* I had hoped that we’d also be able to just write html+css for the text in the header and status but that doesn’t seem to be possible. So this part looks just like it used to.
* Styling of buttons is possible but they lose their native look.

Disadvantages:

* Not all features yet
* Needs more dependencies
* Not sure if it will run everywhere
* We might have to support both Qt and Tk for some time (maybe forever)
* PyQt6 and PySide are competing frameworks. In principle they might both work but this needs a small compatibility layer (mainly for imports)
* We got used to the old, rough look (and on macOS it is actually fine – see below for the faux 3d chin of Pacman)

### Virtualenv Qt6

Virtualenv qt

### Conda Qt6 (disclaimer: I only tested the version that pip would install)
Conda qt

In comparison, these show the differences/problems with Tk:

### Conda Tk (no aliasing)

Conda tk

### Virtualenv Tk (aliasing in text but not in graphics)

Virtualenv tk

### macOS Tk (does aliasing but look at the colour of the buttons)

macOS tk

### Detail (Faux 3d out of nowhere – although this actually looks cool)

Pacman with chin


otizonaizit:
> The possibilities are endless Screenshot from 2023-05-19 14-30-13

that's called _cultural appropriation_ or _racial and ethnic stereotype_, you choose ;-)

Debilski:
I got the idea from a tequila bottle. I’m innocent here :)

PR #757, opened 2023-07-01
error reporting on importing teams sometimes shows wrong/unclear text

otizonaizit:
this is a failing test that reproduces an issue we observed in Mexico City:

Running the code wiht the pelita CLI shows the right error:

```
Could not load test_broken.py: invalid syntax (doesnotexistbelieveme.py, line 1)): SyntaxError ...
[and then a nice traceback with the pointer to the file and the line with the SyntaxError]

```
In the test we get instead a different thing:

`AttributeError: module 'teamy' has no attribute 'move'
`

--> we have observed this "wrong" error reporting also with the CLI, but we can't reproduce it right now. My guess is that we have to go through the whole error-catching on import logic and fix it/streamline it.

### DIFF
--- a/test/test_pelita_player.py
+++ b/test/test_pelita_player.py
@@ -21,6 +21,15 @@ def noteam():
     return None
 """
 
+MODULE_IMPORT_ERROR = """
+import broken
+
+TEAM_NAME = "noteam"
+
+def move(bot, state):
+    return bot.position
+"""
+
 # TODO: The modules should be unloaded after use
 
 class TestLoadFactory:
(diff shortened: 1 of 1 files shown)


Debilski:
I integrated the test in c3bbb8f and right now it works (when the import is made relative).

However: The whole import business is completely brittle. When adding a new test case the module name that is passed to `load_team_from_module` must be different from those in all other tests. If the name was in use before, the results are completely wrong and unexpected.

As a side note: I wasted 45 minutes working on this because my pelita root folder contained a file `broken.py` with no syntax error that got imported instead. It is a mess.

Issue #760, opened 2023-07-01
bot.homezone and bot.walls are sets, bot.food is a list


otizonaizit:
sets have the problem of having a non-deterministic order, so that if the code depends on the order the games are not replicable.

I think that for this kind of things we should always return list of coordinates. I think we used sets for efficiency reasons, but I think easiness of use and students' expectations are more important in this case.

Debilski:
The efficiency discussion was done in #708 . Food was kept as a list in order to keep random.choice working.

Debilski:
Merged in #767

PR #763, opened 2023-07-11
make walls a tuple of tuples on the client's side

otizonaizit:
This makes the usage of walls for client code more predictable and allow for fully reproducible games. Given that bot.walls is actually cached, I don't think there are any relevant implication with respect to speed (see #708).

Fixes: #760


Debilski:
Lgtm. All in all I think we can neglect speed on the client side as long as it is not too slow. (Maybe I’ll do a check for this later.)
If a client needs faster code, they should use their own, better suited representation for the maze anyway (like a graph).

Debilski:
Merged in #767

PR #764, opened 2023-08-01
[DRAFT] Remove setup.{cfg,py} and add pyproject.toml

Debilski:
I chose [flit](https://flit.pypa.io/en/stable/) as a build-system over setuptools because it automatically includes `pelita.__version__` as the version number of the package. Flit also apparently includes *all* files in the repo in the sdist automatically, so there is less configuration needed.

If anyone opines for a more featureful build system, I am happy to have a discussion.


Debilski:
Merging. Tested on several OS and it seemingly installs correctly.

Issue #769, opened 2023-08-31
Use a nicer pelita URL scheme for network bots


Debilski:
Having to type `remote:tcp://10.10.10.1:12345` is very unwieldy and arcane. Using `pelita://10.10.10.1:12345` would be an improvement. (Or we get rid of the URL scheme altogether.)

Additionally, we should err with a better message in case the IP address is not valid (number missing, etc.).

PR #777, opened 2023-10-09
Improved Pelita server mode

### LINKED ISSUE #776: Add zmq protocol to dispatch multiple players with `--remote`
# tl;dr

instead of

    pelita ./my_bot.py remote:tcp://151.11.0.123:12345
    pelita ./my_bot.py remote:tcp://151.11.0.123:12346

we want to do

    pelita ./my_bot.py pelita://some.server.com/attacker_bots
    pelita ./my_bot.py pelita://some.server.com/defender_bots

# Context

We want to play remote games but zeroconf is not available.

# Problem

Usually, we would have to run `pelita-player --remote TEAM ADDRESS` on a server, where `ADDRESS` is usually of the form `tcp://151.11.0.123:12345`. Now arbitrary many teams can connect to this address using the `remote:tcp://...` syntax and will get the chance to play a remote game against the code given in `TEAM`. (See #769 for syntax changes to make this a bit nicer.)

For additional teams that should be run on the remote server, we will have to run `pelita-player` with a different port in `ADDRESS`  (and a different `TEAM` of course).

In networks in which zeroconf is possible, this is not much of a problem, as the URL details will be hidden from the user. When this is not possible, however, remembering the different ports for the different teams is a bit tedious and opaque. Additionally, the firewall on the server has to be changed to let through traffic for each additional team/port.

# Proposal

* We change the handling of the `remote:tcp` (future `pelita`) protocol to have a default port that is used when no port is specified.
* We allow the specification of a `path` to the URL, which will be used to request a specific player from the remote server. This way, we can have one remote service handle all connections to all remote players.

# Possible implementation

pelita already uses a zmq `DEALER`–`ROUTER` dispatch system in remote games. It should be possible to simply add the path component to the request and have some code on the `pelita-player` site to have a proper mapping between a path and a TEAM.

As a bonus, there should be a minimal API to see which TEAMs are available and they should optionally be advertised via zeroconf.

Debilski:
Will still need refinements (that I’m still working on) but already runs and closes #776 and #769.

The short news is you start a Pelita server like this:

    pelita-server remote-server  --address 0.0.0.0 \
            --team ../BayesAvengers.py avengers \
            --team ../trilobots.py trilobots \
            --advertise 10.10.1.2

(--advertise is only needed when we want to use zeroconf)

This will start a server on port 41736 (which is the default for the new URL scheme) and we can run a match with both teams on the server as follows:

    pelita --ascii pelita://10.10.1.2/avengers pelita://10.10.1.2/trilobots

(The specific IP should not be needed and could also be a hostname.)

When the network allows for using zeroconf then it should also be possible to use SCAN and detect all players on the server automatically.

It is not wildly incompatible to the old version but I did not put in any special effort to make it work with the current PyPI version (Pelita 2.3.1), so all clients will need an update in order to play against the new server.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -570,7 +570,8 @@ def prepare_bot_state(game_state, idx=None):
         'enemy': enemy_state,
         'round': game_state['round'],
         'bot_turn': bot_turn,
-        'timeout_length': game_state['timeout_length']
+        'timeout_length': game_state['timeout_length'],
+        'max_rounds': game_state['max_rounds'],
     }
 
     if bot_initialization:
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -228,7 +228,7 @@ def setup_viewers(viewers=None, options=None, print_result=True):
             viewer_state['viewers'].append(ReplayWriter(open(viewer[1], 'w')))
         elif viewer in ('tk', 'tk-no-sync'):
             if not zmq_publisher:
-                zmq_publisher = ZMQPublisher(address='tcp://127.0.0.1:*')
+                zmq_publisher = ZMQPublisher(address='tcp://127.0.0.1')
                 viewer_state['viewers'].append(zmq_publisher)
             if viewer == 'tk':
                 viewer_state['controller'] = setup_controller()
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -506,12 +506,18 @@ def prepare_bot_state(game_state, idx=None):
     """
 
     bot_initialization = game_state.get('turn') is None and idx is not None
+    bot_finalization = game_state.get('turn') is not None and idx is not None
 
     if bot_initialization:
         # We assume that we are in get_initial phase
         turn = idx
         bot_turn = None
         seed = game_state['rnd'].randint(0, sys.maxsize)
+    elif bot_finalization:
+        # Called for remote players in _exit
+        turn = idx
+        bot_turn = None
+        seed = None
     else:
         turn = game_state['turn']
         bot_turn = game_state['turn'] // 2
@@ -977,9 +983,10 @@ def check_exit_remote_teams(game_state):
     """ If the we are gameover, we want the remote teams to shut down. """
     if game_state['gameover']:
         _logger.info("Gameover. Telling teams to exit.")
-        for team in game_state['teams']:
+        for idx, team in enumerate(game_state['teams']):
             try:
-                team._exit()
+                team_game_state = prepare_bot_state(game_state, idx=idx)
+                team._exit(team_game_state)
             except AttributeError:
                 pass
 
--- a/pelita/network.py
+++ b/pelita/network.py
@@ -10,6 +10,10 @@ import zmq
 
 _logger = logging.getLogger(__name__)
 
+# 41736 is the word PELI(T)A when read upside down in reverse without glasses
+# The missing T stands for tcp
+PELITA_PORT = 41736
+
 ## Pelita network data structures
 
 # ControlRequest
--- a/pelita/network.py
+++ b/pelita/network.py
@@ -1,9 +1,9 @@
 
 import json
 import logging
-import re
 import sys
 import time
+from urllib.parse import urlparse
 import uuid
 
 import zmq
@@ -53,42 +53,20 @@ class ZMQClientError(Exception):
 DEAD_CONNECTION_TIMEOUT = 3.0
 
 
-def extract_port_range(address):
-    """ We additionally allow for setting a port range in rectangular brackets:
-        tcp://127.0.0.1:[50100:50120]
-    """
-    range_pattern = re.compile(r"(?P<addr>.*?):\[(?P<port_min>\d+):(?P<port_max>\d+)\]")
-    random_pattern = re.compile(r"(?P<addr>.*?):\*")
-    port_pattern = re.compile(r"(?P<addr>.*?):(?P<port>\d+)")
+def bind_socket(socket: zmq.Socket, address, option_hint=None):
+    parsed_address = urlparse(address)
 
-    m = range_pattern.match(address)
-    if m:
-        return {"addr": m.group(1), "port_min": int(m.group(2)), "port_max": int(m.group(3))}
-    m = random_pattern.match(address)
-    if m:
-        return {"addr": m.group(1), "port_min": None, "port_max": None}
-    m = port_pattern.match(address)
-    if m:
-        return {"addr": address}
-    return {"addr": address}
+    # NB: We cannot use parsed_address.geturl() to generate a nice url for zmq
+    # as this will eat the empty hostname in a file path and zmq does not like that.
+    # file:///tmp/a -> file:/tmp/a
 
-def bind_socket(socket, address, option_hint=None):
     try:
-        address_range = extract_port_range(address)
-        addr = address_range["addr"]
-        try:
-            port_min = address_range["port_min"]
-            port_max = address_range["port_max"]
-            if port_min and port_max:
-                port = socket.bind_to_random_port(addr, port_min, port_max)
-            else:
-                port = socket.bind_to_random_port(addr)
-            bind_addr = "{0}:{1}".format(addr, port)
-            return bind_addr
-
-        except KeyError:
-            socket.bind(addr)
-            return addr
+        if parsed_address.scheme == 'tcp' and parsed_address.port is None:
+            port = socket.bind_to_random_port(address)
+            return f'tcp://{parsed_address.hostname}:{port}'
+        else:
+            socket.bind(address)
+            return address
 
     except (zmq.ZMQError, zmq.ZMQBindError) as e:
         print('error binding to address %s: %s' % (address, e), file=sys.stderr)
@@ -319,7 +297,7 @@ class ZMQPublisher:
 
 
 class Controller:
-    def __init__(self, address='tcp://127.0.0.1:*', zmq_context=None):
+    def __init__(self, address='tcp://127.0.0.1', zmq_context=None):
         self.address = address
         if zmq_context:
             self.context = zmq_context
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -271,9 +271,9 @@ publisher_opt.add_argument('--publish', type=str, metavar='URL', dest='publish_t
                            help=long_help('Publish the game to this zmq socket.'))
 publisher_opt.add_argument('--no-publish', const=False, action='store_const', dest='publish_to',
                            help=long_help('Do not publish.'))
-parser.set_defaults(publish_to="tcp://127.0.0.1:*")
+parser.set_defaults(publish_to="tcp://127.0.0.1")
 
-advanced_settings.add_argument('--controller', type=str, metavar='URL', default="tcp://127.0.0.1:*",
+advanced_settings.add_argument('--controller', type=str, metavar='URL', default="tcp://127.0.0.1",
                                help=long_help('Channel for controlling the game.'))
 advanced_settings.add_argument('--external-controller', const=True, action='store_const',
                                help=long_help('Force control by an external controller.'))
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -8,11 +8,13 @@ import random
 import sys
 
 from rich.console import Console
+from rich.markup import escape
 from rich.prompt import Prompt
 
 import pelita
 from .script_utils import start_logging
 
+from pelita.network import PELITA_PORT
 # TODO: The check_team option
 from pelita.tournament import check_team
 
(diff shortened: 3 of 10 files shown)


Debilski:
> It is not wildly incompatible to the old version but I did not put in any special effort to make it work with the current PyPI version (Pelita 2.3.1), so all clients will need an update in order to play against the new server.

Actually, the old clients do still work (also `SCAN` works) but they cannot select a player with the path element and will always play against the default player on the server.

otizonaizit:
Hey @Debilski : I am running remote games for my students using this branch during the holidays... please don't break it for the time being :)))

otizonaizit:
(including not force-pushing anymore :))) )

otizonaizit:
(by the way: the interface is quite cool, thanks!!!)

Debilski:
This is why they invented personal repositories and branches, you know. :)

But noted.

otizonaizit:
yes, yes, but I want to profit from any improvement you make make on top of this branch without messing up with cherry-picking stuff :))

Debilski:
Ah, ok, so you want me to still push but only the good commits. Got it. :D

otizonaizit:
>Ah, ok, so you want me to still push but only the good commits. Got it. :D

yes, that was the plan :)

Debilski:
@otizonaizit nothing urgent but could you share some of the error messages from your server that were supposedly caused by bots?

otizonaizit:
Yes, but my guess is that they come from port-scanners/bots, not for legitimate bots:
  - type of error a):
  ```
Error JSONDecodeError('Expecting value: line 1 column 1 (char 0)') when parsing incoming message. Ignoring. pelita_server.py:172
  ```

  - type of error b):
  ```
Error UnicodeDecodeError('utf-32-be', b'\x00\x00\x00\x00\x00Cookie: mstshash=Administr\r\n\x01\x00\x08\x00\x03\x00\x00\x00', 4, 8, 'code point not in range(0x110000)') when parsing incoming message. Ignoring. pelita_server.py:172
  ```

  - type of error c):
  ```
Traceback (most recent call last):
  File "/home/tiziano/software/virtualenvs/pelita/bin/pelita-server", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tiziano/git/pelita/pelita/scripts/pelita_server.py", line 425, in remote_server
    server.start()
  File "/home/tiziano/git/pelita/pelita/scripts/pelita_server.py", line 316, in start
    dealer_id, message = self.router_sock.recv_multipart()
    ^^^^^^^^^^^^^^^^^^
ValueError: too many values to unpack (expected 2)
Use --session-key XXXX to for the admin API.  
  ```

Debilski:
Unnecessary for the fix, but I’ve found out why this happens and I will share it here so that future chat bots can help me better next time (looking at you ChatGPT 3.5).

In the version 1 format (there seem to be newer format specs that are a bit more complicated and also I’ve only managed to figure it out for the router type socket but anyway), a zmq message consists of a bunch of *frames* that are specified as octets of:

    LENGTH, FLAGS, FRAME_DATA*

The length is the number of octets of data+1. I’ll get to the flags later. This is a frame:

    echo -e '\x06\x00Hello'

A message that is accepted by a router socket consists of (minimum) two frames. One containing an optional identifier (the empty frame `\x01\x00` will have zmq generate an id for us) from the sender and the data.

Adding a print statement to the pelita server:

        dealer_id, message = self.router_sock.recv_multipart()
        print(dealer_id, message)

and sending the message

    echo -e '\x06\x00Hello\x07\x00Pelita'  | nc localhost 41736

prints

     b'Hello' b'Pelita'
    [11:23:32] Error JSONDecodeError('Expecting value: line 1 column 1 (char 0)')   pelita_server.py:174
               when parsing incoming message. Ignoring.                                                 
    Error JSONDecodeError('Expecting value: line 1 column 1 (char 0)') when parsing incoming message. 
    Ignoring.

Obviously, `b'Pelita'` is not a proper json string so this rightly gets rejected and we found the cause of error type a).

Error type b) is simply sending something that cannot get parsed to unicode correctly.

For error type c) to occur we need to look at the FLAGS octet from the data frame. The logic is: If the final bit is 0 then this is the final frame. If it is set to 1 then we can just add another frame afterwards.

    echo -e '\x06\x00Hello\x07\x01Pelita\x07\x00server'  | nc localhost 41736

This will the code to break as now `self.router_sock.recv_multipart()` receives three messages (which is kind of the only proper bug here as the other error types keep working).

Finally, why do we even see this so often? Malformed messages usually get quietly ignored by zmq, however the rdp format can somehow fit the format and I guess is quite often tried by spam bots:

    echo -e '\x03\x00\x00\x13\x0e\xd0\x00\x00\x00\x00\x00Cookie: mstshash=Admin'


The quick solution would be to ignore (and only trace log) everything that is so obviously malformed. (The only initial message that we accept is a json dict with a `REQUEST` key so there is a lot of messages that we can rightly ignore, but this still means that we have our json parser converting every random bot’s garbage so I’m still thinking about changing the format here slightly.)

Debilski:
The value errors with incoming messages will now only trigger a debug message (which won’t be shown usually).

Debilski:
I experimented with having the server in a systemd service and starting the players as `systemd-run` (all players could the for example share a scope that take as most as x% CPU from the server) but unfortunately this seems to require that pelita-server runs as root (which we might not want). With the server run as root it works fine.

In https://github.com/jupyterhub/systemdspawner/pull/100 they lay out a solution for a similar situation using template unit files + polkit to start them. One option would be to instantiate a template per player (but then for every player there could only be one at a time – unless the players spawn sub-players themselves inside their unit :upside_down_face: ). The other option would have us write startup configuration files on the fly and initiate the templates with the file name. (We need to have a way to pass a communication socket from server to player.)

I sense the need for a pelita for kubernetes. :grimacing:

otizonaizit:
But why not just start with a static list of players read directly from 
the configuration file? Then we just need a normal systemd service. 
Pelita as root is a no-go in my opinion.
For dynamically adding players we can take some time and add the 
functionality later, no?

25 Apr 2024 19:17:53 Rike-Benjamin Schuppner ***@***.***>:

Debilski:
Sure, if you want to run everything (server and players) in one and the same service, then it is not a problem at all. (My post was mostly meant as a future reference point.) I guess I should have made that more clear.

If you don’t care about a rogue player interfering with everything *inside* the service, then it is easily made secure for everyone else *outside*. You can run it as a DynamicUser and harden it almost as much as you like:

```
[Unit]
Description=Pelita server
Documentation=https://github.com/ASPP/pelita
After=network-online.target
Wants=network-online.target


[Service]
WorkingDirectory=/opt/pelita_server
ExecStart=/opt/pelita_server/venv/bin/pelita-server remote-server --address 0.0.0.0 --config /opt/pelita_server/config.yaml

DynamicUser=yes

ProtectHome=yes
ProtectSystem=strict

DevicePolicy=closed
KeyringMode=private
LockPersonality=yes
MemoryDenyWriteExecute=yes
NoNewPrivileges=yes
PrivateDevices=yes
PrivateTmp=yes
PrivateUsers=yes
ProtectClock=yes
ProtectControlGroups=yes
ProtectHostname=yes
ProtectKernelLogs=yes
ProtectKernelModules=yes
ProtectKernelTunables=yes
ProtectProc=invisible
RemoveIPC=yes
RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6 AF_NETLINK
RestrictNamespaces=yes
RestrictRealtime=yes
RestrictSUIDSGID=yes

SystemCallArchitectures=native
SystemCallFilter=

CapabilityBoundingSet=~CAP_CHOWN CAP_FSETID CAP_SETFCAP
CapabilityBoundingSet=~CAP_DAC_OVERRIDE CAP_DAC_READ_SEARCH CAP_FOWNER CAP_IPC_OWNER
CapabilityBoundingSet=~CAP_LINUX_IMMUTABLE CAP_IPC_LOCK CAP_SYS_CHROOT CAP_BLOCK_SUSPEND CAP_LEASE 
CapabilityBoundingSet=~CAP_SETUID CAP_SETGID CAP_SETPCAP CAP_CHOWN CAP_NET_ADMIN 
CapabilityBoundingSet=~CAP_SYS_ADMIN CAP_SYS_BOOT CAP_SYS_PACCT CAP_SYS_PTRACE CAP_SYS_RAWIO CAP_SYS_TIME CAP_SYS_TTY_CONFIG 
CapabilityBoundingSet=~CAP_WAKE_ALARM  CAP_MAC_ADMIN CAP_MAC_OVERRIDE 

[Install]
WantedBy=multi-user.target
```

    > systemd-analyze security pelita-server.service
    → Overall exposure level for pelita-server.service: 3.0 OK 🙂

Debilski:
(I copy-pasted the options together. Not sure what really is needed and what else we could activate. `systemd-analyze security` needs a bisect mode. :) )

otizonaizit:
> [Unit]
> Description=Pelita server
> Documentation=https://github.com/ASPP/pelita
> After=network-online.target
> Wants=network-online.target
>
> [Service]
> WorkingDirectory=/opt/pelita_server
> ExecStart=/opt/pelita_server/venv/bin/pelita-server remote-server --address 0.0.0.0 --config /opt/pelita_server/config.yaml
>
> DynamicUser=yes
>
> ProtectHome=yes
> ProtectSystem=strict
>
> DevicePolicy=closed
> KeyringMode=private
> LockPersonality=yes
> MemoryDenyWriteExecute=yes
> NoNewPrivileges=yes
> PrivateDevices=yes
> PrivateTmp=yes
> PrivateUsers=yes
> ProtectClock=yes
> ProtectControlGroups=yes
> ProtectHostname=yes
> ProtectKernelLogs=yes
> ProtectKernelModules=yes
> ProtectKernelTunables=yes
> ProtectProc=invisible
> RemoveIPC=yes
> RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6 AF_NETLINK
> RestrictNamespaces=yes
> RestrictRealtime=yes
> RestrictSUIDSGID=yes
>
> SystemCallArchitectures=native
> SystemCallFilter=
>
> CapabilityBoundingSet=~CAP_CHOWN CAP_FSETID CAP_SETFCAP
> CapabilityBoundingSet=~CAP_DAC_OVERRIDE CAP_DAC_READ_SEARCH CAP_FOWNER CAP_IPC_OWNER
> CapabilityBoundingSet=~CAP_LINUX_IMMUTABLE CAP_IPC_LOCK CAP_SYS_CHROOT CAP_BLOCK_SUSPEND CAP_LEASE
> CapabilityBoundingSet=~CAP_SETUID CAP_SETGID CAP_SETPCAP CAP_CHOWN CAP_NET_ADMIN
> CapabilityBoundingSet=~CAP_SYS_ADMIN CAP_SYS_BOOT CAP_SYS_PACCT CAP_SYS_PTRACE CAP_SYS_RAWIO CAP_SYS_TIME CAP_SYS_TTY_CONFIG
> CapabilityBoundingSet=~CAP_WAKE_ALARM CAP_MAC_ADMIN CAP_MAC_OVERRIDE
>
> [Install]
> WantedBy=multi-user.target
>
> > systemd-analyze security pelita-server.service
> → Overall exposure level for pelita-server.service: 3.0 OK 🙂
>

wow, that's heavy stuff! I didn't even know you could set all these parameters for a systemd service ;)
But well, if everything works fine with these restrictions, why we don't just use them? We can get rid of redundant ones later. Also, we can ask a systemd developer about it ;-)

Debilski:
I managed to get it to run with socket activation. No we can even turn off the network inside the service. (Score is now 2.6)

otizonaizit:
Hey @Debilski , is this branch ready to merge? It would be cool to have a release latest next week. For me it is important that the communication protocol is fixed. The server code can be refined later, as long as the client running a release can still talk to the server...

Debilski:
I still have a few cleanups that I want to do here (and maybe have some final thoughts about versioning). And there needs to be more documentation. Other than that, I plan to merge latest on Friday and I suppose we can make a release then.
Pelita 2.4 will be compatible to Pelita server 2.4; Pelita 2.5 might not be.

Debilski:
I think there will still be changes in the protocol in future version. If we can live with this, I think we should merge the server now and refine it in the future.

otizonaizit:
yes, merge!

Issue #789, opened 2024-02-28
Add new `Bot.graph` attribute


otizonaizit:
After several years of experience with the new `Bot` API it is clear that almost all players sooner or later use the `pelita.utils.walls_to_graph` function to generate a graph representation of the maze.

In the demo bots we show how to store this into `state`.

Inexperienced programmers tend to overlook the fact that it is expensive to create the graph, and prefer to just do it at every turn instead of trying to understand how to use `state`.

My suggestion is to just bite the bullet and make `Bot.graph` a default attribute for the `Bot` object. It can be created once at the beginning of the game and cached.

This will remove the need of using `state` for something which is actually not different than `bot.walls` in terms of use.

otizonaizit:
Well, we can protect all attributes by making them read-only properties. So they will get an error when they try to do anything with it. Or, for `graph`, we can just use this: https://networkx.org/documentation/stable/reference/generated/networkx.classes.function.freeze.html
I think the problem of mutability is easy to explain in the docs if we can't find a nice fix.

But, by the way, `deepcopy(state)` is science fiction: I have never seen it used in code from our students.

Debilski:
But freezing means that the weights can still change, I think this isn’t sufficient. And people reading the docs is also sci-fi. :)

otizonaizit:
Yes, ok. I mean, let's create the graph at the beginning, attach a copy to `Bot` and if they mess it up they'll find out, and we can point them to the docs if this happens. I still find this better than the current situation, where lots of people are simply regenerating the graph over and over again. And semantically it kind of makes sense to attach it as a static attribut to `Bot`, just like we do with `Bot.walls`.

PR #792, opened 2024-05-15
add a new `graph` cached attribute to the `Bot` object

otizonaizit:
Introduce a new cached attribute `graph` to the `Bot` object.
This PR also moves the `walls_to_graph` function from `utils.py` to `team.py`, where it arguably now belongs. An alias `walls_to_graph` is left in `utils.py` for backward compatibility.

Fixes: #789

### DIFF
--- a/pelita/team.py
+++ b/pelita/team.py
@@ -154,6 +154,9 @@ class Team:
         # Cache the homezone so that we don’t have to create it at each step
         self._homezone = create_homezones(self._shape, self._walls)
 
+        # Cache the graph representation of the maze
+        self._graph = walls_to_graph(self._walls)
+
         return self.team_name
 
     def get_move(self, game_state):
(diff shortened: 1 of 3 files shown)


otizonaizit:
Not sure it makes sense to have the deprecated decorator. New code will not use the old function. Old code is not going to be updated by anyone, so the deprecation won't motivate anyone to update the code. In my opinion we can leave the import there for an indefinite time.

Before merging I want to add some tests to be sure that if a bot changes the graph right now those changes are indeed lost. I am not sure given the way the caching is implemented. Sorry, I didn't think of it when I submitted the PR, otherwise I would have put the WIP label...

We can then use the frozen decorator then, which should pass the still-to-be-written tests, in case the current implementation is indeed broken.

Debilski:
Ok but if we keep it around then we should still flag it so that is is not removed by some import checker. Or remove it now and live with the consequences.

PR #793, opened 2024-05-17
Check that weights can be changed in the graph.

Debilski:
Check that weights are not immutable in a graph view and ensure that both bots see a view to a different cached graph.

### DIFF
--- a/test/test_team.py
+++ b/test/test_team.py
@@ -438,6 +438,42 @@ def test_bot_graph():
     assert state['errors'] == [{}, {}]
     assert state['fatal_errors'] == [[], []]
 
+def test_bot_graph_is_half_mutable():
+    # Test that a bot can change the weights of the graph
+    # Changing the weights affects the future self of the bot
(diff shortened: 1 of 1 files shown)


test/test_team.py
```diff
@@ -438,6 +438,42 @@ def rough_bot(bot, state):
...
+    #  by  #
+    #......#
+    ########
+    """
+    def blue(bot, state):
+        if bot.turn == 0 and bot.round == 1:
+            assert bot.graph[1, 1][1, 2].get('weight') is None
+            bot.graph[1, 1][1, 2]['weight'] = 1
+        else:
+            assert bot.graph[1, 1][1, 2].get('weight') == 1
```
otizonaizit, line 457: wait: from this line it seems that the changed weights are indeed visible to the other bot?

otizonaizit: does changing the weights affect the computation of the shortest path? This could be the source of very interesting bugs :-)

PR #798, opened 2024-05-30
Expose team_time in Bot

### LINKED ISSUE #797: make the cumulative time counter available to clients
The `Bot` object should have an attribute `time` where the cumulative time for the team shown in the GUI is made available. This is useful for example when benchmarking different strategies. There's no way at the moment to programmatically access that value, as far as I could see.
Of course the client could time itself, but this has its own pitfalls. So it's better to just make that value available, even if we all know it's not super precise...

Debilski:
The test could use some more creativity.

Closes #797

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -555,7 +555,8 @@ def prepare_bot_state(game_state, idx=None):
         'bot_was_killed': game_state['bot_was_killed'][own_team::2],
         'error_count': len(game_state['errors'][own_team]),
         'food': list(game_state['food'][own_team]),
-        'name': game_state['team_names'][own_team]
+        'name': game_state['team_names'][own_team],
+        'team_time': game_state['team_time'][own_team]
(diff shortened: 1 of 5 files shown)


Debilski:
wow. Fixing these tests took way longer than expected …

Debilski:
Pedagogic rebase

otizonaizit:
thanks, looks good! I added a label to mark the needs for docs update in [pelita_template](https://github.com/ASPP/pelita_template/). We need to find a better way to mark these dependencies...

Debilski:
It would be nice to have a live stopwatch running in the UI instead of only incrementally updating it, but I believe we cannot really synchronise this with the actual game state.

Debilski:
FWIW Python 3.13 should have a more accurate `time.monotonic()` on Windows. https://github.com/python/cpython/issues/88494

PR #803, opened 2024-07-01
Refactoring of ci_engine

Debilski:
Work in progress. I’ve slightly modernised it but there is a big problem currently with the hashing of the modules that needs to be fixed.

### DIFF
--- a/.github/workflows/test_pytest.yml
+++ b/.github/workflows/test_pytest.yml
@@ -100,3 +100,34 @@ jobs:
       run: |
         pelita-tournament --non-interactive --viewer null
       timeout-minutes: 10
+
+  contrib:
+    # The tournament does not need to run everywhere.
+    # Only test on the latest ubuntu/macos for recent versions of Python.
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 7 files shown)


Debilski:
Hashing problem has been solved with a few lines of code in `pelita-player`. Instead of using the currently-broken `ModuleFinder` or doing it manually with `importlib` and/or `ast` (this approach is tiresome as we have to account for every different way that relative modules can be defined; furthermore we meet the Pelita code base’s old enemy again: You cannot import two scripts from two different locations that both import the module `bot`), we simply ask `pelita-player` to import a game, scan the list of `sys.modules` and generate a hash from all modules relative to the `spec` (ie. the bot script).

The old version (back when it worked) would scan *all* dependencies, so that it would also hash changes that had occurred in the main Pelita code base or somewhere else in Python. However, skipping this also makes it much faster (and it does not matter in 99% of the cases).

Additional changes done to some database queries and tables should be a bit better when using a lot of players. The config format has not been changed yet and no special error handling has been added (however, the final_state is now recorded and can be queried with sqlite json functions which helps a lot with debugging already).

CLI additions:

    python contrib/ci_engine.py --print  # just prints the ranking

    python contrib/ci_engine.py --nohash # runs a game without re-hashing (and potentially resetting) the players

PR #804, opened 2024-07-04
BF: TK UI shows wrong homezone for blue team

otizonaizit:
Fix longstanding bug introduced in 92ca5553f61e05c5edaf9f3bd240628bb1ef06f8

The homezone for the blue team shown in the field status (bottom right) is one square too big. Typical off-by-one error ;-)

(see screenshot)
screenshot1

### DIFF
--- a/pelita/ui/tk_canvas.py
+++ b/pelita/ui/tk_canvas.py
@@ -705,7 +705,7 @@ class TkApplication:
                 has_food = pos in game_state['food']
                 is_wall = pos in game_state['walls']
                 bots = [idx for idx, bot in enumerate(game_state['bots']) if bot==pos]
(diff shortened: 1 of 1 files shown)


otizonaizit:
by the way, the failed check is completely unrelated... any idea why sometimes these checks do get a timeout?

Debilski:
That’s what happens if you don’t have a dedicated function for everything. Thanks.

Debilski:
> by the way, the failed check is completely unrelated... any idea why sometimes these checks do get a timeout?

I assume there is a race condition in the cleanup of a broken Pelita player. Maybe even similar to the one we had in the Pelita server. Looking into it.

Issue #799, opened 2024-06-20
Exception in bot crashes server mode


otizonaizit:
I am reporting here an issue for future reference, because at the moment I can not implement a simple reproducer.

While running in server mode:

```server$ pelita-server remote-server --address XXXXX --port YYYY --config pelita_server_conf.yaml```

if one of the configured bot throws an Exception during a game, it seems like the server keeps on running (systemd reports the service as running) but it stops responding to new remote requests:

```
client$ pelita demo01_stopping.py pelita://XXXXX
Remote player requested. Scanning server for players.
Server did not reply in 5000 ms.
No remote team selected. Exiting.
```

The Exception found in the log was the following one:
```
Traceback (most recent call last):
  File "ZZZZ/pelita_private_bots/sparring_bots/aspp2022_3/bot.py", line 45, in move
    next_pos = check_next_pos( bot, state , next_pos)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "ZZZZ/pelita_private_bots/sparring_bots/aspp2022_3/group3_utils.py", line 42, in check_next_pos
    next_pos = check_legal_status(bot, state, next_pos)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "ZZZZ/pelita_private_bots/sparring_bots/aspp2022_3/group3_utils.py", line 211, in check_legal_status
    next_pos = smart_random_choice( bot, state)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "ZZZZ/pelita_private_bots/sparring_bots/aspp2022_3/group3_utils.py", line 120, in smart_random_choice
    next_pos = bot.random.choice(   bot.legal_positions  )
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/random.py", line 373, in choice
    raise IndexError('Cannot choose from an empty sequence')
IndexError: Cannot choose from an empty sequence
```

I am assuming the Exception and the freeze are connected, but a reproducer would be needed to really make sure that the two things are indeed causally connected :-)

otizonaizit:
OK, at least now I can confirm that the exception freezes the server. For proper debugging we need of course an easier reproducer.

Debilski:
[...]

PR #801, opened 2024-06-24
RF: Add a send queue for outgoing messages to avoid broken players stalling the server

Debilski:
Closes #799 by putting outgoing messages in a send queue.

Shielding individual sends with a timeout is not sensible here as it would stall the server when there are crashed bots.

Draft because I am not sure if this is a good design.

PR #805, opened 2024-07-04
Do not attempt to send an exit message to a team with a fatal error

Debilski:
As was mentioned in https://github.com/ASPP/pelita/pull/804 we sometimes have timeouts in our CI. In this case, it could be tracked down to the test:

    pytest test/test_remote_game.py::test_remote_move_failures[player_move_import_error-False-ModuleNotFoundError]

I cannot reproduce it right now but it is likely that one of two things happens:

The tested player has an import error, so the game setup fails and triggers a fatalerror. This error causes the game to finish and we end up in `def check_exit_remote_teams` where we notify the players of how the game went down and that they should shut down. As the player has already had a fatal error this can

a) cause a hang in zmq.send (similar to https://github.com/ASPP/pelita/issues/799); this should actually have been taken care of with our helper class `ZMQConnection` but it will still induce a one second timeout.

b) cause a hang in RemotePlayer.__del__ when the process is terminated or maybe cause it in this method’s call to exit

Since we do not need to send anything to a player with a fatalerror (after all, we should assume that it cannot receive any more messages), we simply stop sending a message.

Coincidentally, this would also be enough to fix https://github.com/ASPP/pelita/issues/799 but our fix in https://github.com/ASPP/pelita/pull/801 is still more generic and appropriate.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -988,6 +988,9 @@ def check_exit_remote_teams(game_state):
         for idx, team in enumerate(game_state['teams']):
             if len(game_state['fatal_errors'][idx]) > 0:
                 _logger.info(f"Not sending exit to team {idx} which had a fatal error.")
+                # We pretend we already send the exit message, otherwise
+                # the team’s __del__ method will do it once more.
+                team._sent_exit = True
                 continue
             try:
                 team_game_state = prepare_bot_state(game_state, idx=idx)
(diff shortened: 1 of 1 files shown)


Debilski:
Nice side effect: running the tests in test_remote_game.py is now 10 seconds faster.

PR #812, opened 2024-08-02
New and enlarged collection of layouts

otizonaizit:
This PR updates the old layouts and extends the collection so that we now have 1000 normal sized layouts. Strategies that require manually checking every layout should be much less feasible now. And it was time to update the layouts so we don't get bored after 5 years of using the same ones ;)

I added a couple of shell scripts for making it easier next time to generate the layouts. The big layouts are quite heavy to generate. To get all 1200 new layouts it took around 3.5 hours on my laptop.


Debilski:
Looks about right. That will show those students who want to sight-spot the best squatting location. :)

PR #821, opened 2024-09-27
Reduce default shadow distance to 1

otizonaizit:
In Heraklion we noticed that the value 3 for the shadow distance for food relocation was a bit too broad. It made a purely attacking strategy overwhelmingly more worth investing than a defending or mixed strategy. by reducing the shadow to 1 we still avoid pure camping, but still make it worth to explore defending strategies.

I know it all sounds quite hand-wavy, but it is very difficult to put numbers onto this reasoning.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -37,7 +37,7 @@ NOISE_RADIUS = 5
 MAX_FOOD_AGE = 30
 
 #: Food pellet shadow distance
-SHADOW_DISTANCE = 3
+SHADOW_DISTANCE = 1
 
 class TkViewer:
     def __init__(self, *, address, controller, geometry=None, delay=None, stop_after=None, fullscreen=False):
--- a/test/test_filter_gamestates.py
+++ b/test/test_filter_gamestates.py
@@ -9,7 +9,7 @@ from pelita.game import setup_game, play_turn, prepare_bot_state, split_food
 from pelita.layout import parse_layout
 from pelita.player import stepping_player
 import pelita.utils
-
+import pelita.game
 
 def make_gamestate():
     def dummy_team(bot, state):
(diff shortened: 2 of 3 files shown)


Debilski:
I suppose we can do that but I disagree that from `n=5` teams in Heraklion we can conclude that the choice of radius 3 was a failure.
It would be neat to have a way of not having to hard-code this into Pelita itself (and then change it before each course that uses Pelita and for which a different value should be tested).

otizonaizit:
> I suppose we can do that but I disagree that from `n=5` teams in Heraklion we can conclude that the choice of radius 3 was a failure. It would be neat to have a way of not having to hard-code this into Pelita itself (and then change it before each course that uses Pelita and for which a different value should be tested).

Yes, I am with you with this. But we would then need to re-think a bit the way we distribute pelita. This stuff should be a cfg file. For each course we would distribute a specific cfg file, maybe even in `pelita_template`.

I don't think it is a lot of work, but I think it's fair to postpone it.
Even if `n=5`, Pietro tested a lot while writing his new bot, and it became pretty clear that raiuds 3 makes defending too hard.

Debilski:
Yeah, a file in pelita_template might be feasible but it becomes tricky when pelita is run from a different folder. And installing it to a `.config` folder is too opaque in my opinion. This would only work if there is some course-related branding (‘ASPP 2024’ or a color scheme) added to the config file as well, that is also shown and makes it obvious in the GUI which ruleset is loaded.

 It *could* become an attribute of the layout and not of the game. (I’ll add some more to that idea in #825.)

Debilski:
It might make more sense (for the problem we want to solve) to have the food shadow radius stop at walls. But I suppose this introduces yet another concept.

otizonaizit:
It may or it may not. That's exactly the reason why I want to make the 
change and test this particular change live for a semester course :)

30 Sept 2024 19:35:51 Rike-Benjamin Schuppner ***@***.***>:

Debilski:
No, no, it 100% makes more sense to stop the food shadow radius at a wall and only count empty cells in maze space. There is absolutely no need to test this. :)
We introduced it as a defender penalty but there is no reason to have that penalty if the defender cannot reach a food item. The only argument here is the question which shadow structure is easier to explain.

Of course, with a radius of 1, there is no difference in implementation or effect. :)

otizonaizit:
> No, no, it 100% makes more sense to stop the food shadow radius at a wall and only count empty cells in maze space. There is absolutely no need to test this. :)

But this is exactly my point. I **want** to test this. And the only way to test this, as we had to learn from the change done in v2.5.0, is to implement the change and see what students do with it.

Debilski (commented):
Merging

PR #823, opened 2024-09-27
Make it possible to generate mazes with dead ends and chambers

otizonaizit:
This PR does not change the default generation algorithm, so mazes are still generated without dead-ends and chambers by default.

### DIFF
--- a/pelita/maze_generator.py
+++ b/pelita/maze_generator.py
@@ -378,7 +378,7 @@ def add_pacmen(maze):
     maze[1, -2] = b'y'
     maze[2, -2] = b'x'
 
-def get_new_maze(height, width, nfood, seed=None):
+def get_new_maze(height, width, nfood, seed=None, dead_ends=False):
     """Create a new maze in text format.
 
     The maze is created with a recursive creation algorithm. The maze part of
@@ -391,6 +391,12 @@ def get_new_maze(height, width, nfood, seed=None):
     height, width -- the size of the maze, including the outer walls
     nfood -- number of food dots for each team
     seed -- if not None, the random seed used to generate the maze
+    dead_ends -- if True allow for dead ends and chambers in the maze
+
+    A dead-end is a node with connectivity one.
+    A chamber is a sub-graph such that there is a node in the sub-graph, the
+    entrance to the chamber, that when removed from the graph will result in the
+    graph to be split into two disconnected graphs.
     """
     if width%2 != 0:
         raise ValueError(f'Width must be even ({width} given)')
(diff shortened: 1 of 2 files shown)


Debilski:
I suppose this one is uncontroversial :D
Maybe we should also add a test for this in `test_maze_generation.py`? (Btw. I think that line with `* iteration` is kind of useless?)

```
@pytest.mark.parametrize('iteration', range(1,11))
def test_get_new_maze(set_seed, iteration):
    # generate a few mazes and check them for consistency
    local_seed = random.randint(1,2**31-1)*iteration
    maze_str = mg.get_new_maze(8,16,nfood=15,seed=local_seed)
```

otizonaizit:
> I suppose this one is uncontroversial :D Maybe we should also add a test for this in `test_maze_generation.py`?

theoretically yes, but not sure what to test. The maze generation works in a way that a maze is generated and it may or may not contain dead ends (it is not predictable). Then it removes the dead ends. So testing that we can generate a maze with dead ends is kind of useless.

> (Btw. I think that line with `* iteration` is kind of useless?)
> 
> ```
> @pytest.mark.parametrize('iteration', range(1,11))
> def test_get_new_maze(set_seed, iteration):
>     # generate a few mazes and check them for consistency
>     local_seed = random.randint(1,2**31-1)*iteration
>     maze_str = mg.get_new_maze(8,16,nfood=15,seed=local_seed)
> ```

Why? I think the test is generating 10 mazes and verify that they are consistent and are properly formatted. It is a kind of integration test. It could be done only once, but doing it 10 times doesn't hurt either, no? Or are you worried about the test being slow?

Debilski:
> > I suppose this one is uncontroversial :D Maybe we should also add a test for this in test_maze_generation.py?
> 
> theoretically yes, but not sure what to test. The maze generation works in a way that a maze is generated and it may or may not contain dead ends (it is not predictable). Then it removes the dead ends. So testing that we can generate a maze with dead ends is kind of useless.

Yeah, I copied the other test and realised this. But I thought maybe you had a better idea. :)


I mean this line specifically. Why make it a multiple of `iteration`?

    local_seed = random.randint(1, 2**31-1) * iteration

otizonaizit:
> > > I suppose this one is uncontroversial :D Maybe we should also add a test for this in test_maze_generation.py?
> > 
> > 
> > theoretically yes, but not sure what to test. The maze generation works in a way that a maze is generated and it may or may not contain dead ends (it is not predictable). Then it removes the dead ends. So testing that we can generate a maze with dead ends is kind of useless.
> 
> Yeah, I copied the other test and realised this. But I thought maybe you had a better idea. :)
> 
> I mean this line specifically. Why make it a multiple of `iteration`?
> 
> ```
> local_seed = random.randint(1, 2**31-1) * iteration
> ```

to make it more random ;-)

Debilski (commented):
LGTM

PR #824, opened 2024-09-27
Add a collection of layouts with dead ends

otizonaizit:
Depends on #823 

This is posted as a separate PR, but should be merged once pelita has been changed to make use of the layouts in a user-configurable way. If merged as-is it would start using the new layouts right away.


Debilski:
I will do my main commenting in #825 but … do we really need that many layouts? There may be some merit in having fewer layouts, as students may then spend more time looking at the specifics on how some mazes look and specifically write code for them. With too many mazes this is not feasible (at least not during ASPP) and students may not do it at all. (I’m not speaking about hard-coding the maze but with fewer mazes one may find certain patterns in them.)

Of course, we could always have a restriction for a specific course so no big deal.

otizonaizit:
> Do we really need that many layouts? There may be some merit in having fewer layouts, as students may then spend more time looking at the specifics on how some mazes look and specifically write code for them. With too many mazes this is not feasible (at least not during ASPP) and students may not do it at all. (I’m not speaking about hard-coding the maze but with fewer mazes one may find certain patterns in them.)
> 
> Of course, we could always have a restriction for a specific course so no big deal.

That is a way to look at it. Another way is that many layouts make the game more unpredictable and as a consequence more fun. Sure there must be a sweet spot between 100 layouts – too few, as students already shown that it can be exploited – and 1000 layouts – no one even thought of exploiting it –. But I am not sure it is worth exploring it.

The real thing should be to just generate the maze on the fly. Unfortunately our maze generator is too slow for it. But I'd rather spend time in improving the maze generator to be usable live instead of finding the **right** number of layouts. Then we would even infinite number of mazes :-)))

Debilski:
> Then we would have an infinite number of mazes :-)))

If all mazes are possible then I want this to be the last-chance final (will probably break the student bots as there is more than 30 food :/ )

image

Debilski (approved):

PR #825, opened 2024-09-30
Use dead ends

otizonaizit:
Use the newly defined layouts with dead ends. This PR also sets a default proportion of 0.25 for the layouts with dead ends, which means that every 4th game will have a layout with dead ends.

This PR should be merged after #823 and #824

### DIFF
--- a/pelita/layout.py
+++ b/pelita/layout.py
@@ -1,5 +1,6 @@
 import importlib.resources as importlib_resources
 import io
+import os
 import random
 
 # bot to index conversion
@@ -74,13 +75,13 @@ def get_available_layouts(size='normal', dead_ends=False):
         size = ''
 
     av_layouts = []
-    for resource in importlib_resources.files('pelita._layouts').iterdir():
-        if resource.is_file() and resource.name.endswith('.layout') and size in resource.name:
-            layout_name = resource.name.removesuffix('.layout')
-            if dead_ends and 'dead_ends' in resource.name:
-                av_layouts.append(layout_name)
-            if not dead_ends and 'dead_ends' not in resource.name:
-                av_layouts.append(layout_name)
+    for file in os.listdir(importlib_resources.files('pelita._layouts')):
+        if dead_ends:
+            cond = file.endswith('.layout') and size in file and 'dead_ends' in file
+        else:
+            cond = file.endswith('.layout') and size in file and 'dead_ends' not in file
+        if cond:
+            av_layouts.append(file.removesuffix('.layout'))
 
     return sorted(av_layouts)
 
--- a/pelita/layout.py
+++ b/pelita/layout.py
@@ -9,7 +9,7 @@ BOT_I2N = {0: 'a', 2: 'b', 1: 'x', 3: 'y'}
 
 RNG = random.Random()
 
-def get_random_layout(size='normal', seed=None, dead_ends=False):
+def get_random_layout(size='normal', seed=None, dead_ends=0):
     """ Return a random layout string from the available ones.
 
     Parameters
@@ -23,9 +23,9 @@ def get_random_layout(size='normal', seed=None, dead_ends=False):
         'big'    -> width=64, height=32, food=60
         'all'    -> all of the above
 
-    dead_ends: bool
-        if set, return a layout from the collection with dead_ends, otherwise
-        return a layout without dead_ends
+    dead_ends: float
+        Return a layout from the collection with dead ends with probabilty dead_ends.
+        By default never return a layout with dead_ends.
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


pelita/layout.py
```diff
@@ -74,13 +75,13 @@ def get_available_layouts(size='normal', dead_ends=False):
...
 
     av_layouts = []
-    for resource in importlib_resources.files('pelita._layouts').iterdir():
-        if resource.is_file() and resource.name.endswith('.layout') and size in resource.name:
-            layout_name = resource.name.removesuffix('.layout')
-            if dead_ends and 'dead_ends' in resource.name:
-                av_layouts.append(layout_name)
-            if not dead_ends and 'dead_ends' not in resource.name:
-                av_layouts.append(layout_name)
+    for file in os.listdir(importlib_resources.files('pelita._layouts')):
```
Debilski, line 78: Very interesting. I think part of the speed gain comes from skipping the `is_file` check but even when narrowing it down with `glob` I cannot get the same speed as with `listdir`. It does lead me to believe, however, that we should do the `get_available_layouts` differently.

pelita/utils.py
```diff
@@ -20,7 +20,7 @@ def _parse_layout_arg(*, layout=None, food=None, bots=None, seed=None):
 
     # prepare layout argument to be passed to pelita.game.run_game
     if layout is None:
-        layout_name, layout_str = get_random_layout(size='normal', seed=seed)
+        layout_name, layout_str = get_random_layout(size='normal', seed=seed, dead_ends=DEAD_ENDS)
```
Debilski, line 23: I think `def get_random_layout` should have the dead_end factor already as a default (just as `size="normal"` is a default).

pelita/layout.py
otizonaizit: I spent seom time trying to understanding it too. I think the `pathlib` is just slow. And I don't see an easy way of changing the code much, because we need `importlib.resources`. If we use the `.contents` attribute it is even faster, and it returns just a list of strings. Unfortunately that attribute is deprecated and this leaves us with only `.files`... `glob` and `iterdir` are both slower, even if `glob` is a bit better.
We could decide to hardcode the layout names and just assume that the files are going to be there. `get_available_layouts` doesn't need to be dynamic as it is now and can file at runtime if then for some absurd reasons the files are not there.

pelita/utils.py
otizonaizit: Yes, but I thought until we are still testing this allows us to only change `game.DEAD_ENDS`. Hiding it as an argument to `get_random_layout` seems odd to me for a change that is so significant. But we could change it after we have battle tested it and we know that it is going to stay.

Debilski:
Some ideas to the layout loading mechanism. I think with 2500 layouts we are approaching a point at which we do not want to traverse the filesystem every time that Pelita loads.

Moreover, with the re-addition of dead ends, we have now another attribute that needs to be selected on startup (and currently not all files with dead_end in their name are guaranteed to have dead ends).
Let us think about adding a database (file with a list of json objects should do), semi-automatically generated and consistency-guaranteed through pytest, that enumerates all layouts and records all attributes that can be found for them:

    [ ...,
    {
      "name": "dead_ends_normal_001.layout",
      "shape": [32, 16],
      "size": "normal",
      "dead_ends": True,
      "num_dead_ends": 8,
      "num_chambers": 12,
      "food_age_radius": 1
    },
    ...]

This could also allow us to add additional properties, like having a different food age radius **depending on the layout**. (Maybe increase it on all dead_end layouts.) Not sure we need to expose everything and make it selectable to the users of Pelita but for fine-tuning purposes it might still be useful to know how many layouts we have with more than 12 chambers or something like that.


To take this idea further: The problem of having different settings/layouts per course (mentioned in #821) could be solved by shipping the layouts separately from Pelita (maybe in a course-specific pelita_template). This could either be a config file that simply restricts the number of layouts (and sets up some defaults) that are loaded from the main game. Or it would need to come with a whole layouts folder.


(Somewhat unsolved: Selecting in a 1:4 ratio no-dead_ends/dead_ends will still need to be hard coded.)

otizonaizit:
While I agree with what you say, I think this discussion is more of a 
long term discussion. For the PR here the real added cost is on the order 
of some tens of milliseconds for the second call of pelota, i.e. after 
the operating system I/O caches have been filled.

Let's move the discussion to a new issue and discuss there to find a 
better long term solution.

30 Sept 2024 19:38:05 Rike-Benjamin Schuppner ***@***.***>:

otizonaizit:
For the long term discussion also relevant: it is not at all obvious what 
it means to count dead ends. It is easy to find nodes with connectivity 
1. But this would count a "tunnel" of length 5 as 5 dead ends, where 
instead it may make more sense to count it as 1.

1 Oct 2024 09:32:37 Tiziano Zito ***@***.***>:

Debilski:
> While I agree with what you say, I think this discussion is more of a long term discussion. For the PR here the real added cost is on the order of some tens of milliseconds for the second call of pelota, i.e. after the operating system I/O caches have been filled. Let's move the discussion to a new issue and discuss there to find a better long term solution.

Yes, *after* the I/O caches have been filled. Chances are that the removal of the `is_file` check really improved the initial startup speed at least on network drives (starting Pelita did a `newfsstatat` syscall on every layout) but maybe also on Windows (though I don’t know how `is_file` is implemented there). So it is accidentally actually becoming faster with this PR merged. :)

Debilski (approved):
Good for now. We’ll fine-tune it later.

PR #843, opened 2024-11-13
NF: Add --stop-after-kill option

### LINKED ISSUE #841: Add UX option `--stop-after-kill`
This could be a CLI option and/or a Tk shortcut/button. The effect would be that the game pauses every time that a bot gets killed.

The feature would allow a user to investigate whatever debugging output was written to the command line right before a bot was killed, without this being drowned in messages from further moves.

Debilski:
Closes #841


otizonaizit:
LGTM!

Issue #844, opened 2024-11-13
Revert bot state. Bring back simulate move


Debilski:
We currently prepare the bot state in `game.py` in a scheme that mirrors the way it is later used by the `Bot` API.

```yaml
game_state:
    team: 
        team_index: 1
        bot_positions: [[29,1],[30,2]]
        score: 0
        kills: [0,0]
        deaths: [0,0]
        bot_was_killed: [false,false]
        error_count: 0
        food: [[...]]
        name: Smart Random Players
        team_time: 0.001439541985746473
    enemy: 
        team_index: 0
        bot_positions: [[1,8],[1,13]]
        is_noisy: [true,true]
        score: 0
        kills: [0,0]
        deaths: [0,0]
        bot_was_killed: [false,false]
        error_count: 0
        food: [[...]]
        shaded_food: []
        name: Smart Eating Players
        team_time: 0.006159459124319255
    round: 2
    bot_turn: 1
    timeout_length: 3
    max_rounds: 300
```

There is some sense to this as it leaves out unnecessary information. For example, we do not need to send `is_noisy` information for our own bot.

This comes with a disadvantage, however, in that more advanced code that might want to use a different API has to revert these decisions. Specifically, code that wants to re-use the `apply_move` function to simulate a move has to reimplement the logic of `apply_move` targeted at the `Bot` API or write a conversion function from `Bot` API back to `game_state` (and fear that this breaks whenever new properties are added).

My proposal would be to create the `Bot` API on the client side and instead send a data structure similar to the game state (properly cleaned up from unnecessary/confidential data). A normal user would not see any difference when using the `Bot` API but a power user could still access the original game state behind a attribute.

Reference to a prior implementation of `simulate_move`: https://github.com/ASPP/pelita/pull/399

otizonaizit:
In principle it sounds as a good idea, especially if it can be done without changing the `Bot` API. My guess is taht the devil will be in the details of the implementation, but it is worth a try ;)

PR #851, opened 2025-03-06
RF: Use classic game_state for communication with the bots

Debilski:
This undos the split in team/enemy in the state that a team receives.

The reasoning is that this format makes it easier to reuse functions from game.py in one’s own code.

Works and checks all tests but will need some cleanup and full implementation of #844 to be really useful.


otizonaizit:
Hey, I don't want to give a premature review, but I am a bit worried by the direction this is taking.

I appreciated a lot the idea of cleaning up a bit the use of the game state and rationalizing it's usage thorough the code base. But now I see types creeping in everywhere and I think this is a bigger endeavor and a contentious one at that.
Except for helping the IDEs to give better completions, I am not sure what all the new type definitions are doing for the code base. It's not like we have some form of static type checking going on in the CI or so.
But more importantly, this makes the code harder to read and harder to work with, because now you have another thing to worry about when you look at the code and it becomes harder and harder to edit the code without a properly configured  IDE. Especially if we want to keep pelita approachable by relative inexperience programmers and increase the bus factor to a number much higher than 1, I think it is of paramount importance to have easy to understand code at the cost of less efficiency, and also at the cost of less consistency.

This is a big discussion, so maybe this is not the right forum to have it.

Debilski:
> Except for helping the IDEs to give better completions

Exactly. I am using these to catch errors during development with the help of my (pretty much unconfigured) IDE. Apart from that, these are WIP commits in a draft PR. ;)

Main motivation though: Why put things in a comment that no-one reads like in  https://github.com/ASPP/pelita/blob/main/pelita/network.py#L17-L32 :

```
## Pelita network data structures

# ControlRequest
# {__action__}

# ViewerUpdate
# {__action__, __data__}

# Request
# {__uuid__, __action__, __data__}

# Reply
# {__uuid__, __return__}

# Error
# {__uuid__, __error__, __error_msg__}
```

when we can use modern Python features for the same thing? In that sense, yes, I am planning to rewrite these comments to `TypedDicts`. Whether we use them to actually annotate functions is another story.

otizonaizit:
ok, ok, then I'll retreat and wait for the WIP to be finalized.

PR #852, opened 2025-03-06
WIP: Add layout database

jbdyn:
This PR adds a Python script for generating the database for Pelita mazes.

### DIFF
--- a/pelita/create_db.py
+++ b/pelita/create_db.py
@@ -1,0 +1,203 @@
+import json
+
+import jsbeautifier
+import networkx as nx
+from rich.pretty import pprint
+from rich.progress import track
+
+from pelita.layout import get_available_layouts, get_layout_by_name, parse_layout
+from pelita.maze_generator import find_chamber
+
+
+def position_in_maze(pos, shape):
+    x, y = pos
+    w, h = shape
+
+    return 0 <= x < w and 0 <= y < h
+
+
+def walls_to_graph(walls, shape):
+    w, h = shape
+    directions = [(1, 0), (0, 1), (-1, 0), (0, -1)]
+
+    graph = nx.Graph()
+    # define nodes for maze
+    for x in range(w):
+        for y in range(h):
+            if (x, y) not in walls:
+                # this is a free position, get its neighbors
+                for delta_x, delta_y in directions:
+                    neighbor = (x + delta_x, y + delta_y)
+                    # we don't need to check for getting neighbors out of the maze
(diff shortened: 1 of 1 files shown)


Debilski:
I refined the existing `walls_to_graph` implementation in #853 so that it can be reused here.

jbdyn:
> I refined the existing `walls_to_graph` implementation in https://github.com/ASPP/pelita/pull/853 so that it can be reused here.

Oh, sorry, I didn't see that. I also reworked the code in `create_db.walls_to_graph`.

otizonaizit:
> > I refined the existing `walls_to_graph` implementation in #853 so that it can be reused here.
> 
> Oh, sorry, I didn't see that. I also reworked the code in `create_db.walls_to_graph`.

do you have a profiling for this? How much faster is it to start with a fully connected graph and remove the nodes?

Debilski:
I just profiled it. @jbdyn’s new variant is twice as fast as ours on the normal mazes. Interestingly, `nx.grid_2d_graph` does not apply any magic at all (if fills the whole thing with Python for loops), so this is probably a case of simply choosing the right iterator for the right data structure. (see: https://networkx.org/documentation/stable/_modules/networkx/generators/lattice.html#grid_2d_graph )

Debilski:
```python

def find_chambers(G: nx.Graph, shape):
    w, h = shape

    main_chamber = set()
    chamber_tiles = set()

    for chamber in nx.biconnected_components(G):
        max_x = max(chamber, key=lambda n: n[0])[0]
        min_x = min(chamber, key=lambda n: n[0])[0]
        if min_x < w // 2 <= max_x:
            # only the main chamber covers both sides
            main_chamber = set(chamber)
            continue
        else:
            chamber_tiles.update(set(chamber))

    # remove shared articulation points with the main chamber
    chamber_tiles -= main_chamber

    # combine connected subgraphs
    subgraphs = G.subgraph(chamber_tiles)
    chambers = list(nx.connected_components(subgraphs))

    return chambers, chamber_tiles
```

Using this instead of `paint_chambers` reduces the time from 9.3s to 6.5s on my system. (We can now also skip searching for the cuts before.)

otizonaizit:
@jbdyn : is it OK if we push to this branch to move it forward?

jbdyn:
> is it OK if we push to this branch to move it forward?

Sure!

otizonaizit:
Obsoleted by #856 and #867 : how nice that we don't need this one anymore :-))) Thanks @jbdyn !

PR #853, opened 2025-03-07
ENH: walls_to_graph takes a shape argument for optimisation and other speed-ups

Debilski:


otizonaizit:
There are more optimisations that cut down execution time to almost half in #852 . Maybe you could take them over here so that we profit from them also in `team.py`?

otizonaizit:
...or I can do it, as you prefer.

Debilski:
The absolutely fastest way would be to have `parse_layout` return `nx.Graph` directly, I suppose. Not sure if it is worth it for normal Pelita code but I suppose the maze analysis functions could profit from this.

Debilski:
Found the issue in our code: I forgot that we made `walls` a list which slowed down the `in` checks massively and makes it as fast as @jbdyn’s variant. And we can then again half the time by only looking at the positive neighbours.

Benchmarks:

```
from pelita.team import walls_to_graph
from pelita.layout import get_available_layouts, get_layout_by_name, parse_layout
import networkx as nx

for layout_name in get_available_layouts(size='normal', dead_ends=True):
    layout_str = get_layout_by_name(layout_name)
    layout = parse_layout(layout_str)

    g = walls_to_graph(layout['walls'], layout['shape'])
```

    Version with walls in a list and all neighbours: 4s
    Version with walls in a set and only positive neighbours: 0.5s
    Adding all to nx and removing walls: 1s


Variant 3 for reference:

```
def walls_to_graph(walls, shape):
    graph = nx.grid_2d_graph(*shape)
    graph.remove_nodes_from(walls)
    return graph
```

otizonaizit:
> Found the issue in our code: I forgot that we made `walls` a list which slowed down the `in` checks massively and reverting this makes it as fast as @jbdyn’s variant. And we can then again half the time by only looking at the positive neighbours.
> 
> Benchmarks:
> 
> ```
> from pelita.team import walls_to_graph
> from pelita.layout import get_available_layouts, get_layout_by_name, parse_layout
> import networkx as nx
> 
> for layout_name in get_available_layouts(size='normal', dead_ends=True):
>     layout_str = get_layout_by_name(layout_name)
>     layout = parse_layout(layout_str)
> 
>     g = walls_to_graph(layout['walls'], layout['shape'])
> ```
> 
> ```
> Version with walls in a list and all neighbours: 4s
> Version with walls in a set and only positive neighbours: 0.5s
> Adding all to nx and removing walls: 1s
> ```
> 
> Variant 3 for reference:
> 
> ```
> def walls_to_graph(walls, shape):
>     graph = nx.grid_2d_graph(*shape)
>     graph.remove_nodes_from(walls)
>     return graph
> ```

But given that the walls are coming from json, we would have to pay the price of conversion to set anyway in `team.py`, no? Anyway, if it is worth it, we can first convert to set, then generate the graph, then convert to list for the bot API.

Debilski:
Good point. It arrives as a list anyway so we can just convert it one time in our `walls_to_graph` and profit from faster access. No other code needs to be touched. (Which is what is implemented now.)

Debilski:
But the time win on the client side is not really why we need to optimise this. As the benchmark is per 1000 matches, this is 4ms vs 0.5ms during set_initial.
It gets more important on the server side.

Issue #854, opened 2025-03-07
Add better Tk *viewer* mode


Debilski:
The basic idea is:

- Start a Tk viewer with `pelita-tkviewer tcp://127.0.0.1:12345`
- In another window either
  - run `pelita --publish-to tcp://127.0.0.1:12345 --null --delay 50` to start a game (with an optional fixed minimum delay time; but this could also be handled on the Tk viewer side)
  - run proper Pelita code that uses something like: `viewer = Publisher('tcp://127.0.0.1:12345'); viewer.update(game_state)`
- The viewer should now print the situation

This should also work with layouts.

Optionally:
- Give `pelita-tkviewer` an option to hide the controls.
- Add `--only-show-layout` option to `pelita` that will not start a game. (We could also use `--stop-at 0` for this but this would fail, for example, if the layout is incomplete – currently forbidden but might be useful for debugging.)

#819 would make things a bit nicer as it should improve the state handling mechanism but it is not strictly needed.

Debilski:
Additionally, the viewer should accept an optional attribute in the maze spec:

    walls: [(0,0), ...]
    shape: (32, 16)
    ...
    overlays:
    - {"fill": "#2200ff", "pos": [[1,1], [1,2]]}
    - {"fill": "#22ffff", "pos": [[3,1], [3,2], [3,3]]}

Which would (in this example) colour the given fields.

Issue #849, opened 2025-03-06
introduce a database of mazes with descriptive metadata, for example number and length of dead-ends, etc.


otizonaizit:

otizonaizit:
Looking at #852 , we need to discuss what to store in the layout database. This depends in my opinion on how we plan to use that information. I envision something like this:

```bash
pelita --trapped-food 0.3 bot1.py bot2.py
```

where `--trapped-food` represent the (maximum) proportion of food which is located within dead-ends, tunnels and/or chambers. This would replace the current `--dead-ends` switch.

The way it would work should be the following:

  - if a layout is explicitly specified and the layout contains no food, food is distributed randomly but according to `--trapped-food` in a "soft" way, i.e. if the layout does not have enough trapped squares it won't error our but just do that best it can
  - if layout is explicitly specified and the layout already contains food, a warning is issued and the `--trapped-food ` switch is ignored
  - if layout is **not** explicitly specified, a random layout is selected among those in the database  that allow us to satisfy `--trapped-food` and food is distributed randomly according to `--trapped-food`. For this to work, the layouts will be stored without food in the database.
  
What do you think @Debilski ? If we agree on this use case, we can design the database to make this the privileged and optimized use case. Given that the database is private, I think we can change it later with not too much trouble should the need arise, given that #852 is optimized as hell and tuns in ~10 seconds on 1600 layouts.

Debilski:
Couple of thoughts:

* I suppose this would mean that we also get rid of the mazes with/without dead ends distinction? Mazes can have all the dead ends/chambers that they like, because who cares if we do not put food in them?
* Without the need for this distinction and with an algorithm that finds all chambers in a maze in 1/160 seconds, we might not need a maze database at all (at least not yet). We use the `--trapped-food` (name feels a bit weird) option and the food is arranged according to specification.
* Not having a db will make life a little easier. (Less to think about, fewer moving parts to maintain.)
* The food relocation algorithm will need to be aware of the desired trapped food ratio.
* I think mazes that contain food should (/could) also be reshuffled. For development situations, it could be easier, if the food in a maze does not reshuffle each time one uses it, but this can simply be done by not using the `--trapped-food` flag.
* For a tournament, the food should absolutely be reshuffled. (Although I do not see any problem with *not* reshuffling it. I kind of doubt that there is any real danger in knowing the food locations in advance.)
[...]

PR #856, opened 2025-03-10
Optimize Maze Generation

jbdyn:
Hey guys,

analogous to #852, I also tried to use the first version of the optimized chamber finding algorithm in the maze generation as well.

The timings went from several seconds down to less than half a second consistently for script execution.
However, looking at the profiling, only 5% (25 ms) are actually maze generation and the rest of the time was spent in importing:

<details>
<summary>absolute times</summary>

```
  _     ._   __/__   _ _  _  _ _/_   Recorded: 09:47:34  Samples:  233
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.480     CPU time: 1.285
/   _/                      v5.0.1

Program: scripts/pelita_createlayout.py

0.479 <module>  pelita_createlayout.py:1
├─ 0.299 <module>  ../__init__.py:1
│  └─ 0.297 <module>  ../game.py:1
│     ├─ 0.183 <module>  ../team.py:1
│     │  └─ 0.181 <module>  networkx/__init__.py:1
│     │        [18 frames hidden]  networkx, importlib
│     ├─ 0.074 <module>  ../viewer.py:1
│     │  ├─ 0.062 <module>  rich/console.py:1
│     │  │     [11 frames hidden]  rich, fractions
│     │  └─ 0.010 <module>  rich/progress.py:1
│     ├─ 0.030 <module>  ../network.py:1
│     │  └─ 0.028 <module>  zmq/__init__.py:1
│     │        [10 frames hidden]  zmq, importlib, enum
│     └─ 0.006 <module>  logging/__init__.py:1
├─ 0.146 <module>  ../maze_generator.py:1
│  └─ 0.146 <module>  numpy/__init__.py:1
│        [30 frames hidden]  numpy, typing
├─ 0.024 main  pelita_createlayout.py:65
│  └─ 0.022 get_new_maze  ../maze_generator.py:381
│     ├─ 0.012 remove_all_chambers  ../maze_generator.py:321
│     │  └─ 0.006 articulation_points  networkx/algorithms/components/biconnected.py:263
│     │     └─ 0.006 _biconnected_dfs  networkx/algorithms/components/biconnected.py:338
│     └─ 0.010 remove_all_dead_ends  ../maze_generator.py:256
│        └─ 0.008 walls_to_graph  ../maze_generator.py:193
└─ 0.008 compile  <built-in>
```
</details>

<details>
<summary>relative times</summary>

```
  _     ._   __/__   _ _  _  _ _/_   Recorded: 09:47:34  Samples:  233
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.480     CPU time: 1.285
/   _/                      v5.0.1

Program: scripts/pelita_createlayout.py

100.0% <module>  pelita_createlayout.py:1
├─ 62.4% <module>  ../__init__.py:1
│  └─ 62.0% <module>  ../game.py:1
│     ├─ 38.2% <module>  ../team.py:1
│     │  └─ 37.8% <module>  networkx/__init__.py:1
│     │        [18 frames hidden]  networkx, importlib
│     ├─ 15.4% <module>  ../viewer.py:1
│     │  ├─ 12.9% <module>  rich/console.py:1
│     │  │     [11 frames hidden]  rich, fractions
│     │  └─ 2.1% <module>  rich/progress.py:1
│     ├─ 6.3% <module>  ../network.py:1
│     │  └─ 5.8% <module>  zmq/__init__.py:1
│     │        [10 frames hidden]  zmq, importlib, enum
│     └─ 1.3% <module>  logging/__init__.py:1
├─ 30.5% <module>  ../maze_generator.py:1
│  └─ 30.5% <module>  numpy/__init__.py:1
│        [30 frames hidden]  numpy, typing
├─ 5.0% main  pelita_createlayout.py:65
│  └─ 4.6% get_new_maze  ../maze_generator.py:381
│     ├─ 2.5% remove_all_chambers  ../maze_generator.py:321
│     │  └─ 1.3% articulation_points  networkx/algorithms/components/biconnected.py:263
│     │     └─ 1.3% _biconnected_dfs  networkx/algorithms/components/biconnected.py:338
│     └─ 2.1% remove_all_dead_ends  ../maze_generator.py:256
│        └─ 1.7% walls_to_graph  ../maze_generator.py:193
└─ 1.7% compile  <built-in>
```
</details>

Would that be fast enough to generate mazes on-the-fly?

For now, I think this would not make the layout database obsolete as one still does not have direct control over the number of dead ends and chambers.

### DIFF
--- a/pelita/maze_generator.py
+++ b/pelita/maze_generator.py
@@ -25,6 +25,7 @@ import numpy as np
 
 
 from .base_utils import default_rng
+from .team import walls_to_graph as walls_to_graph_team
 
 north = (0, -1)
 south = (0, 1)
@@ -312,10 +313,13 @@ def remove_all_chambers(maze, rng=None):
         maze_graph = walls_to_graph(maze)
         # this will find one of the chambers, if there is any
         # entrance, chamber = find_chamber(maze_graph)
-        chambers, chamber_tiles = find_chambers(maze_graph, width)
+        chamber_tiles = find_chambers(maze_graph, width)
 
-        if not chambers:
+        if len(chamber_tiles) == 0:
             break
+
+        subgraphs = maze_graph.subgraph(chamber_tiles)
+        chambers = list(nx.connected_components(subgraphs))
 
         for chamber in chambers:
             # get all the walls around the chamber
@@ -447,24 +451,14 @@ def find_chambers(G: nx.Graph, width: int):
     chamber_tiles -= main_chamber
 
     # combine connected subgraphs
-    subgraphs = G.subgraph(chamber_tiles)
-    chambers = list(nx.connected_components(subgraphs))
+    #subgraphs = G.subgraph(chamber_tiles)
+    #chambers = list(nx.connected_components(subgraphs))
 
-    return chambers, chamber_tiles
-
-
-def sample_nodes(nodes, k, rng=None):
-    rng = default_rng(rng)
-
-    if k < len(nodes):
-        return set(rng.sample(sorted(nodes), k=k))
-    else:
-        return nodes
+    #return chambers, chamber_tiles
+    return chamber_tiles
 
 
 def distribute_food(all_tiles, chamber_tiles, trapped_food, total_food, rng=None):
-    # TODO: ensure that total_food <= len(graph.nodes)
-
     rng = default_rng(rng)
 
     if trapped_food > total_food:
@@ -495,10 +489,172 @@ def distribute_food(all_tiles, chamber_tiles, trapped_food, total_food, rng=None
     else:
         leftover_food_pos = set()
 
-    return sorted(tf_pos | ff_pos | leftover_food_pos)
+    return tf_pos | ff_pos | leftover_food_pos
 
 
-def create_layout(trapped_food, total_food, width, height, rng=None):
+
+def generate_walls(partition, walls, ngaps, vertical, rng=None):
+    rng = default_rng(rng)
+
+    (xmin, ymin), (xmax, ymax) = partition
+
+    # the size of the maze partition we work on
+    width = xmax - xmin + 1
+    height = ymax - ymin + 1
+
+    # if the partition is too small, stop
+    if height < 3 and width < 3:
+        return set()
+
+    # insert a wall only if there is some space in the around it in the
+    # orthogonal direction, i.e.:
+    # if the wall is vertical, then the relevant length is the width
+    # if the wall is horizontal, then the relevant length is the height
+    partition_length = width if vertical else height
+    if partition_length < rng.randint(3, 5):
+        return set()
+
+    # the raw/column to put the horizontal/vertical wall on
+    # the position is calculated starting from the left/top of the maze partition
+    # and then a random offset is added -> the resulting raw/column must not
+    # exceed the available length
+    pos = xmin if vertical else ymin
+    pos += rng.randint(1, partition_length - 2)
+
+    # the maximum length of the wall is the space we have in the same direction
+    # of the wall in the partition, i.e.
+    # if the wall is vertical, the maximum length is the height
+    # if the wall is horizontal, the maximum length is the width
+    max_length = height if vertical else width
+
+    # We can start with a full wall, but we want to make sure that we do not
+    # block the entrances to this partition. The entrances are
+    # - the tile before the beginning of this wall [entrance] and
+    # - the tile after the end of this wall [exit]
+    # if entrance or exit are _not_ walls, then the wall must leave the neighboring
+    # tiles also empty, i.e. the wall must be shortened accordingly
+    if vertical:
+        entrance_before = (pos, ymin - 1)
+        entrance_after = (pos, ymin + max_length)
+        begin = 0 if entrance_before in walls else 1
+        end = max_length if entrance_after in walls else max_length-1
+        wall = {(pos, ymin+y) for y in range(begin, end)}
+    else:
+        entrance_before = (xmin - 1, pos)
+        entrance_after = (xmin + max_length, pos)
+        begin = 0 if entrance_before in walls else 1
+        end = max_length if entrance_after in walls else max_length-1
+        wall = {(xmin+x, pos) for x in range(begin, end)}
+
+    # place the requested number of gaps in the otherwise full wall
+    # these gaps are indices in the direction of the wall, i.e.
+    # x if horizontal and y if vertical
+    # TODO: when we drop compatibility with numpy, this can be more easily done
+    # by just sampling ngaps out of the full wall set, i.e.
+    # gaps = rng.sample(wall, k=ngaps)
+    # for gap in gaps:
+    #     wall.remove(gap)
+    ngaps = max(1, ngaps)
+    wall_pos = list(range(max_length))
+    rng.shuffle(wall_pos)
+    gaps_pos = wall_pos[:ngaps]
+    for gap in wall_pos[:ngaps]:
+        if vertical:
+            wall.discard((pos, ymin+gap))
+        else:
+            wall.discard((xmin+gap, pos))
+
+    # collect this wall into the global wall set
+    walls |= wall
+
+    # define the two new partitions of the maze generated by this wall
+    # these are the parts of the maze to the left/right of a vertical wall
+    # or the top/bottom of a horizontal wall
+    if vertical:
+        partitions = [((xmin,  ymin), (pos-1, ymax)),
+                      ((pos+1, ymin), (xmax,  ymax))]
+    else:
+        partitions = [((xmin,  ymin), (xmax, pos-1)),
+                      ((xmin, pos+1), (xmax,  ymax))]
+
+    for partition in partitions:
+        walls |= generate_walls(
+            partition, walls, max(1, ngaps // 2), not vertical, rng=rng
+        )
+
+    return walls
+
+def generate_full_maze(width, height, ngaps_center, bots_pos, rng=None):
+    # use binary space partitioning
+    rng = default_rng(rng)
+
+    # outer walls are top, bottom, left and right edge
+    walls = {(x, 0) for x in range(width)} | \
+            {(x, height-1) for x in range(width)} | \
+            {(0, y) for y in range(height)} | \
+            {(width-1, y) for y in range(height)}
+
+    # Generate a wall with gaps at the border between the two homezones
+    # in the left side of the maze
+
+    # TODO: when we decide to break backward compatibility with the numpy version
+    # of create maze, this part can be delegated directly to generate_walls and
+    # then we need to rewrite mirror to mirror a set of coordinates around the center
+    # by discarding the lower part of the border
+
+    # Let us start with a full wall at the left side of the border
+    x_wall = width//2 - 1
+    wall = {(x_wall, y) for y in range(1, height - 1)}
+
+    # possible locations for gaps
+    # these gaps need to be symmetric around the center
+    # TODO: when we decide to break compatibility with the numpy version of
+    # create_maze we can rewrite this. See generate_walls for an example
+    ymax = (height - 2) // 2
+    candidates = list(range(ymax))
+    rng.shuffle(candidates)
+
+    for gap in candidates[:ngaps_center//2]:
+        wall.remove((x_wall, gap+1))
+        wall.remove((x_wall, ymax*2 - gap))
+
+    walls |= wall
+    partition = ((1, 1), (x_wall - 1, ymax * 2))
+
+
+    walls = generate_walls(
+        partition,
+        walls,
+        ngaps_center // 2,
+        vertical=False,
+        rng=rng,
+    )
+
+    # make space for the pacmen:
+    for bot in bots_pos:
+        if bot in walls:
+            walls.remove(bot)
+
+    walls = mirror(walls, width, height)
+    return walls
+
+
+def mirror(nodes, width, height):
+    nodes = set(nodes)
+    other = set((width - 1 - x, height - 1 - y) for x, y in nodes)
+    return nodes | other
+
+
+def sample_nodes(nodes, k, rng=None):
+    rng = default_rng(rng)
+
+    if k < len(nodes):
+        return set(rng.sample(sorted(nodes), k=k))
+    else:
+        return nodes
+
+
+def create_maze_graph(trapped_food=10, total_food=30, width=32, height=16, rng=None):
     if width % 2 != 0:
         raise ValueError(f"Width must be even ({width} given)")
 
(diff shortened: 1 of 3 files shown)


jbdyn:
Related discussion about layout database design: #849

otizonaizit:
that is quite impressive. Did you verify that the new algorithm and the old one generate exactly the same maze if started with the same random seed?

otizonaizit:
Half a second is too much to be run at every game, but if we don't require to remove dead ends and chambers anymore, the whole thing would be even faster, no? And in that case we could relax our requirements. We could hard code that we want to "trap" a maximum of 33% of food pellets in chambers/dead-ends, and then do our best depending on how many "trapped" tile we have available on the fly.

jbdyn:
> Did you verify that the new algorithm and the old one generate exactly the same maze if started with the same random seed?

No, but for that I would need to align `pelita_createlayout.py` and `maze_generator.py` as the first one takes a `seed` kwarg and the second one does not, but instead a `rng` object.
Have not looked into this yet.

> Half a second is too much to be run at every game

Okay, but most of this time is importing pelita modules which are not required at all in the maze generation.
This is due to the root `__init__.py`.

> Given the failing test it seems to me that the change also changes the generated maze.

The tests fail because I removed `find_chamber` and introduced a new `find_chambers` (note the `s`).

otizonaizit:
> Half a second is too much to be run at every game, but if we don't require to remove dead ends and chambers anymore, the whole thing would be even faster, no? And in that case we could relax our requirements. We could hard code that we want to "trap" a maximum of 33% of food pellets in chambers/dead-ends, and then do our best depending on how many "trapped" tile we have available on the fly.

@Debilski : what do you think of this approach? I would try together with @jbdyn to implement it on Wednesday in a monster PR. Plan:
  - maze generation on the fly
  - configurable proportion of "trapped" food, intended as a best effort: if there are not enough trapped tiles, then less food will be trapped than requested
  - remove the functionality to remove chambers/dead-ends (could be kept somewhere commented out so taht we don't have to re-invent the wheel in the future

Debilski:
@jbdyn Could you rebase this branch on the updated main? This will make comparisons easier.

Debilski:
Let me see if I can fix #854 before Wednesday as this is probably useful for testing (although shell redirection will already do the job well).

Can we be certain though that the same maze will be generated everywhere? And even if this is the case, I see a few UX problems:

* Currently, we have 1000 mazes and it is always clear which one is used.
* This makes it less overwhelming for the students than having billions of mazes and they can pick any maze and easily discuss it.
* They can also easily copy and paste it and maybe modify it to test something specific. A maze is always in a file (or in a string).
* If we want to change the 1000 mazes, we can simply recreate new ones and ship them.

Cons in the new approach:

* Live-generated mazes can only have a seed as an id. The seed must be short enough to be useful.
* If students want to modify a maze, they will have to pipe it into a file or use special Python code to read and save it.
* If we want to change which maze is generated from `--seed 1`, we have to hard-code a salt value into Pelita (which is used to generate the real seed) and change that *in code*.

Not unsolvable and maybe not too bad (I dislike the salt here, though. Maybe you have a better idea.) but I wanted to note these things.

jbdyn:
> Could you rebase this branch on the updated main? This will make comparisons easier.

@Debilski Done.

otizonaizit:
> Can we be certain though that the same maze will be generated everywhere? And even if this is the case, I see a few UX problems:
> 
>     * Currently, we have 1000 mazes and it is always clear which one is used.
> 
>     * This makes it less overwhelming for the students than having billions of mazes and they can pick any maze and easily discuss it.

in my experience everyone has always assumed that the mazes are generated at run time. The team that asked about it was very surprised when I said that they are indeed pre-generated and that was the moment when they chose to hardcode behavior dependent on the layout name. We decided to have 1000 mazes to make it impossible to do this. Having mazes generated at run time will not surprise anyone, as far as my experience tells me. 

>  * They can also easily copy and paste it and maybe modify it to test something specific. A maze is always in a file (or in a string).

But that is still possible:

```bash
pelita --seed XXX --dump-layout /tmp/my.layout
```
With `--seed` you replay on the same maze with the same food. With `--dump-layout` you save to a string, modify the layout at will and reload later with

```bash
pelita --layout /tmp/my.layout
```

Given that `my.layout` will contain the food, you will be able to play on exactly the situation you want to have. You have full control.

>     * If we want to change the 1000 mazes, we can simply recreate new ones and ship them.

But hey, there's not need to do it if maze generation is fast.

 
> Cons in the new approach:
> 
>     * Live-generated mazes can only have a seed as an id. The seed must be short enough to be useful.

You lost me here. Why does a maze need an id? 

>     * If students want to modify a maze, they will have to pipe it into a file or use special Python code to read and save it.

See the `--dump-layout` option above

>     * If we want to change which maze is generated from `--seed 1`, we have to hard-code a salt value into Pelita (which is used to generate the real seed) and change that _in code_.

I am even more lost here. Why do we care about `--seed 1`?

Debilski:
> in my experience everyone has always assumed that the mazes are generated at run time. The team that asked about it was very surprised when I said that they are indeed pre-generated and that was the moment when they chose to hardcode behavior dependent on the layout name. We decided to have 1000 mazes to make it impossible to do this. Having mazes generated at run time will not surprise anyone, as far as my experience tells me.

Survivorship bias? Maybe teams that were not confused about the layouts never asked? :upside_down_face: 



The problem I am describing boils down to: How do two separate groups in a team communicate which layout they should use for testing.

The first group notices a problem and tells the other group that they should compare it with their own implementation or whatever.

Currently: The first group has the UI open, sees the layout name there and communicates this to the second group.

With auto-generated layouts, they must scroll back to find the seed on the command line (buried in lines of debugging output). And then they must recite something like 20 numbers to the other group so that they can use this as a seed to have the same layout. (And this assumes that this is actually stable between different computers.)

My suggested solution here was to generate a short id from the main seed that is given to the layout generator. This id would be shown in the UI and could be used for communication. (The second group would for example use `--layout 123` to access the layout that the first group sees.) The drawback here is that all layouts with nice seeds are then fixed forever (or until the maze-generating algorithm changes), hence my suggestion to add a salt. But I am not a fan of this idea either.

Obviously people can save and send around layout files, but this makes things quite a bit more involved compared to what we had before.

---

Some thinking later:

What we could do is suggest to the teams that they pre-generate their own set of mazes for themselves in their group repo and whenever pelita is run with `pelita --layout ./folder`, it will draw a random layout from that folder. Then they can decide on their own naming scheme.

Potential breakage, though: There will be this one team who does this and generates only three mazes and then losing because they made some hard-coded assumptions about these mazes.

Debilski:
I start to like the pregen idea. If we had a command that does this automatically, we could tell the teams to use it on the first day or so:

    > pelita-gen-layout --count 1000 ./layout_folder
    Generating 1000 layouts in ./layout_folder. Re-run with --seed 1244546245
    .............done.

And the layout_folder then contains `./layout_folder/0000.layout` or something like that.

otizonaizit:
Really, I never, never see anyone being obsessed about replaying on the same maze. When this happened, it was always connected to reusing the very same seed, i.e. replaying the very same game. I have seen seeds saved in files and communicated through git. Never layout names. Even in the last TU course, a group had a collection of seeds to explore because of problems with the corresponding games. Having the same layout but a different seed seems to me a very exotic debugging configuration (which is of course still achievable, just not by using an id shortcut).

Debilski:
> I have seen seeds saved in files and communicated through git. Never layout names.

That’s my point, though. There is no need to write down a layout name or put it into git, they would just shout it over the table.

But I was just listing the differences that I see. If you think they can be neglected then I am fine with that.

pelita/maze_generator.py
```diff
@@ -202,7 +201,7 @@ def _add_wall(maze, ngaps, vertical, rng):
             _add_wall(sub_maze, max(1, ngaps // 2), not vertical, rng=rng)
 
 
-def maze_to_graph(maze):
+def walls_to_graph(maze):
```
Debilski: Maybe we could make it a bit more explicit (in name, docs and/or type annotation) which functions take a numpy maze and which take a list of wall positions?

otizonaizit: yes, yes, definitely. This is at the moment a moving target, we are changing and adapting the code a lot. It is really a big WIP. I'll ask here for review as soon as the code has stabilized a bit. There are all sorts of horrors in it at the moment :-)

Debilski: Yeah, I thought so. I wish Github had a feature that made these comments private until the review is ready. I wrote it so *I* won’t forget. :)

jbdyn:
Minor detail: `generate_half_maze` was actually supposed to take `nodes` as its first argument, but in `time_graph.py` I passed a `nx.Graph` object. It seems to work either way, and profiling shows no different results.

jbdyn:
We should rename `generate_half_maze` to `generate_walls` though and include flipping the walls as well.

jbdyn:
For the record, the algorithm generating the inner walls is actually called *binary space partitioning*.

Debilski:
Great work!

otizonaizit:
I have refactored the graph-based maze generation completely, adding comments everywhere to make it hopefully more understandable to the future reader. I have added also tests to verify that we can reproduce exactly the numpy-based generated mazes.

The profiling timings are promising (times are for 2000 mazes of normal size):
  - graph-based: 5.5s
  - numpy with dead ends (no layout parsing): 0.9s
  - numpy with dead ends (with layout parsing): 1.2s
  - numpy no dead ends (no layout parsing): 26.9s
  - numpy no dead ends (with layout parsing): 27.1s

So, yes, we are almost 5x slower than the numpy (with layout parsing) with a maze with dead ends. But the timing for the graph-based include the time spent to distribute the food with a fixed proportion in and out of dead-ends. If we wanted to do the same with the old numpy based maze, we would be facing the slower timings. In this sense we could say that we are 5x faster ;-)
And now the code is much more readable.

If we decide to drop numpy compatibility some parts can be refactored even further to make it even more readable. It doesn't work right now because refactoring would imply changing the usage of the random generator so that we wouldn't generate the same maze as numpy with the same seed.

otizonaizit:
It is still WIP, but it is now possible to start looking at it ;-)

otizonaizit:
Integration in pelita CLI is still missing completely.

Debilski:
> I have refactored the graph-based maze generation completely, adding comments everywhere to make it hopefully more understandable to the future reader.

The inline comments are certainly helpful but as a future reader, I would suggest adding comments that explain the purpose of a function as well. :)

For example:

    def generate_walls(partition, walls, ngaps, vertical, rng=None):

Why does it take walls as an argument? Isn’t this supposed to be creating these walls? (I think for what I think it does it should also have a different name that does not include ‘generate’. More like `place_walls` or `partition_maze`. But you decide, I am only 80% sure of what it does.)


Concerning future additions to the food algorithm: Right now the whole thing is (more or less) only concerned with the absolute number of chamber tiles, right? Will it become much more costly once we might decide to distinguish here between chambers, tunnels and possibly also the distance to an entrance?

Debilski:
> Integration in pelita CLI is still missing completely.

If we do live-integration into Pelita, we might want to have different error types for input errors like odd number of tiles for x and ‘could not place enough food. please try again with a new seed’ (or do the latter one automatically).

otizonaizit:
> > I have refactored the graph-based maze generation completely, adding comments everywhere to make it hopefully more understandable to the future reader.
> 
> The inline comments are certainly helpful but as a future reader, I would suggest adding comments that explain the purpose of a function as well. :)

yes, yes, it's in my TODO.

> For example:
> 
> ```
> def generate_walls(partition, walls, ngaps, vertical, rng=None):
> ```
> 
> Why does it take walls as an argument? Isn’t this supposed to be creating these walls? (I think for what I think it does it should also have a different name that does not include ‘generate’. More like `place_walls` or `partition_maze`. But you decide, I am only 80% sure of what it does.)

Agreed.
 
> Concerning future additions to the food algorithm: Right now the whole thing is (more or less) only concerned with the absolute number of chamber tiles, right? Will it become much more costly once we might decide to distinguish here between chambers, tunnels and possibly also the distance to an entrance?

It will be a bit slower, but not dramatically so. The food distribution is super fast. Most of the time is spent in finding the biconnected components. If we go back to distinguish tunnels and chambers it will be maybe twice as slow. But still, we are talking about about 3s over 2000 mazes, so nothing that the user will notice while using pelita.

otizonaizit:
by working on only the left half of the maze for all graph operations I could cut the execution time by almost half, so now to generate 2000 mazes with detection of chambers and distribution of food according to configuration we are down to 3.5s.

Until here the whole thing is generating exactly the same mazes we were generating with numpy when using the same random seed, and I have added tests to verify that this is the case.

Now the question: if we allow for a divergence from numpy we could simplify the code even further (we are talking about 20 lines less), which would help the future reader a lot. The problem is that it may be tricky to verify that the changes are really only efficiency/cosmetics and not really breaking the algorithm. Would you be satisfied if I use as a measure the number of tiles in chambers @Debilski @jbdyn ?

I could generate 100_000 mazes with the old algorithm, 100_000 mazes with the modified one, and then count how many mazes have N tiles in chambers and check that the histograms are compatible? Or can you come up with a better metric that does not require to write huge amount of new code?

otizonaizit:
Or is it better to move this to a new PR?

Debilski:
I guess I’d say if it looks like a maze it does not matter if it is strictly compatible with the old version or not.

What is the difference to the numpy version, though?

otizonaizit:
> ```python
>     gaps = rng.sample(wall, k=max(1, ngaps))                                           
>     for gap in gaps:                                                           
>          wall.remove(gap)
> ```
> 

I also have the hard-core version (untested):

```python
    wall = wall.difference(rng.sample(wall, k=max(1, ngaps)))
```
which is maybe a little too terse ;-)

Debilski:
Maybe you could rewrite the numpy first to also use `sample` (on a `list(range(len(...))`), then regenerate all test mazes and then work again on the non-numpy version and see if this matches with the then-new mazes? Or are the versions also incompatible in the number of items in the list etc?

jbdyn:
@otizonaizit

> Would you be satisfied if I use as a measure the number of tiles in chambers @Debilski @jbdyn ?

I suspect that this measure will be helpful. Since it is about deciding whether the mazes are satisfactory or not, I would have just eyeballed it. We already experienced that our intuition is quite good about a "good" - or better, a "bad" - maze. :smirk:
Here, we can be just artists and free to adapt the code to make the mazes look "satisfactory", regardless of the previous state.

> Or is it better to move this to a new PR?

Just push it here. It is still about optimizing the maze generation. :slightly_smiling_face: 

> Maybe you could rewrite the numpy first

I feel that this would be too much.

Personally, I don't see any difference between

```python
rng.shuffle(wall)                                                        
gaps = wall[:ngaps]
```
and
```python
gaps = rng.sample(wall, k=ngaps)
```

So, IMO just go for the rewrite. :+1:

jbdyn:
> Personally, I don't see any difference between
> 
> ```python
> rng.shuffle(wall)                                                        
> gaps = wall[:ngaps]
> ```
> 
> and
> 
> ```python
> gaps = rng.sample(wall, k=ngaps)
> ```

Let `x` be our population to be `shuffle`d or `sample`d.

[rng.shuffle](https://github.com/python/cpython/blob/2b5feb1f9ea41ccbdc79a8dcd737c3481dc69102/Lib/random.py#L358-L361)

```python
for i in reversed(range(1, len(x))):
    # pick an element in x[: i + 1] with which to exchange x[i]
    j = randbelow(i + 1)
    x[i], x[j] = x[j], x[i]
```

[rng.sample](https://github.com/python/cpython/blob/2b5feb1f9ea41ccbdc79a8dcd737c3481dc69102/Lib/random.py#L442-L446)

```python
n = len(x)
result = [None] * k
pool = list(x)
for i in range(k):
    j = randbelow(n - i)
    result[i] = pool[j]
    pool[j] = pool[n - i - 1]  # move non-selected item into vacancy
```
I generated some walls:

<details>
<summary>shuffled</summary>

`╶╴╶╴████╶╴██`
`██╶╴╶╴╶╴████`
`████╶╴██╶╴╶╴`
`╶╴██████╶╴╶╴`
`████╶╴╶╴██╶╴`
`╶╴╶╴╶╴██████`
`╶╴██╶╴██╶╴██`
`██╶╴╶╴╶╴████`
`██████╶╴╶╴╶╴`
`╶╴╶╴╶╴██████`
`████╶╴╶╴██╶╴`
`██████╶╴╶╴╶╴`
`████╶╴╶╴██╶╴`
`██╶╴╶╴╶╴████`
`╶╴╶╴██╶╴████`
`██╶╴████╶╴╶╴`
`██╶╴╶╴╶╴████`
`██╶╴╶╴████╶╴`
`██╶╴╶╴██╶╴██`
`╶╴██╶╴╶╴████`
`██╶╴██╶╴██╶╴`
`╶╴╶╴██████╶╴`
`╶╴╶╴██████╶╴`
`██╶╴╶╴╶╴████`
`╶╴╶╴████╶╴██`
`╶╴████╶╴╶╴██`
`╶╴██╶╴██╶╴██`
`╶╴████╶╴╶╴██`
`██╶╴██╶╴╶╴██`
`╶╴╶╴██╶╴████`
`╶╴██╶╴██╶╴██`
`╶╴████╶╴╶╴██`
`╶╴██╶╴████╶╴`
`██████╶╴╶╴╶╴`
`██╶╴╶╴╶╴████`
`██╶╴╶╴██╶╴██`
`██╶╴████╶╴╶╴`
`╶╴████╶╴██╶╴`
`██╶╴██╶╴██╶╴`
`██╶╴████╶╴╶╴`
`╶╴████╶╴╶╴██`
`████╶╴██╶╴╶╴`
`╶╴╶╴██╶╴████`
`██╶╴╶╴██╶╴██`
`██╶╴╶╴╶╴████`
`████╶╴██╶╴╶╴`
`╶╴╶╴████╶╴██`
`██╶╴████╶╴╶╴`
`╶╴╶╴██╶╴████`
`╶╴██╶╴╶╴████`

</details>

<details>
<summary>sampled</summary>

`██████╶╴╶╴╶╴`
`████╶╴╶╴╶╴██`
`██╶╴██╶╴╶╴██`
`████╶╴██╶╴╶╴`
`╶╴████╶╴╶╴██`
`██╶╴╶╴████╶╴`
`██╶╴██╶╴╶╴██`
`╶╴╶╴████╶╴██`
`██╶╴╶╴╶╴████`
`██╶╴╶╴████╶╴`
`╶╴╶╴██╶╴████`
`╶╴████╶╴╶╴██`
`╶╴██╶╴╶╴████`
`╶╴╶╴████╶╴██`
`██████╶╴╶╴╶╴`
`██╶╴██╶╴██╶╴`
`████╶╴╶╴╶╴██`
`██████╶╴╶╴╶╴`
`██╶╴╶╴██╶╴██`
`████╶╴██╶╴╶╴`
`╶╴████╶╴██╶╴`
`████╶╴╶╴██╶╴`
`╶╴╶╴██╶╴████`
`╶╴╶╴██████╶╴`
`██╶╴████╶╴╶╴`
`╶╴██╶╴╶╴████`
`╶╴╶╴██╶╴████`
`██╶╴╶╴╶╴████`
`████╶╴██╶╴╶╴`
`██╶╴╶╴████╶╴`
`██████╶╴╶╴╶╴`
`██╶╴╶╴████╶╴`
`████╶╴╶╴██╶╴`
`╶╴██╶╴██╶╴██`
`╶╴████╶╴██╶╴`
`████╶╴╶╴╶╴██`
`████╶╴╶╴╶╴██`
`██╶╴╶╴████╶╴`
`██╶╴████╶╴╶╴`
`██████╶╴╶╴╶╴`
`██╶╴██╶╴╶╴██`
`████╶╴╶╴╶╴██`
`╶╴██████╶╴╶╴`
`╶╴██╶╴╶╴████`
`████╶╴╶╴╶╴██`
`╶╴╶╴████╶╴██`
`╶╴████╶╴╶╴██`
`██╶╴╶╴╶╴████`
`████╶╴██╶╴╶╴`
`██╶╴╶╴████╶╴`

</details>

<details>
<summary>sample_walls.py</summary>

```python
from random import Random

NWALLS = 6
NGAPS = NWALLS // 2

rng = Random()


def get_walls(gaps):
    walls = ["██"] * NWALLS
    for gap in gaps:
        walls[gap] = "╶╴"
    return "".join(walls)


with open("shuffled.md", "w+") as fshuffle, open("sampled.md", "w+") as fsample:
    for i in range(50):
        shuffle_x = list(range(NWALLS))
        sample_x = list(range(NWALLS))

        rng.shuffle(shuffle_x)
        shuffled_gaps = shuffle_x[:NGAPS]
        sampled_gaps = rng.sample(sample_x, k=NGAPS)

        print(f"`{get_walls(shuffled_gaps)}`", file=fshuffle)
        print(f"`{get_walls(sampled_gaps)}`", file=fsample)
```
</details>

otizonaizit:
> ########  ########
> #                #
> #            #   #
> #      #  ####   #
> #   ####  #      #
> #   #...         #
> #   ....         #
> ########  ########

I tested it this way: generate 1000 mazes with the code that works on full mazes, including distribution of food. Generate 1000 mazes with the code that works on half mazes. Verify that the food is distributed in exactly the same way. The problem we had does not happen anymore because I cut in half + 1 column of the right border. I think.
But I can test this layout and see if it works.

otizonaizit:
> I tested it this way: generate 1000 mazes with the code that works on full mazes, including distribution of food. Generate 1000 mazes with the code that works on half mazes. Verify that the food is distributed in exactly the same way. The problem we had does not happen anymore because I cut in half + 1 column of the right border. I think. But I can test this layout and see if it works.

I tested it again. The way I implemented it, it always works because the right side of the maze has no walls except for the external ones. This way any wall that maybe located at the left border can be always walked around going to the right side. The example you posted correctly detects no chambers if you delete all the walls on the right side.
I added a comment in the code so that we don't forget this.

jbdyn:
> it always works because the right side of the maze has no walls except for the external ones

Ouh, _that's_ brilliant! Super neat that this works now with half the effort. :star_struck:

jbdyn:
It just hit me: `shuffle` and `sample` _are indeed_ equivalent, and I can prove it:

This is the definition of the routine in `rng.sample`, [as stated above](https://github.com/ASPP/pelita/pull/856#issuecomment-2740654998).

```python
n = len(x)
result = [None] * k
pool = list(x)
for i in range(k):
    j = randbelow(n - i)
    result[i] = pool[j]
    pool[j] = pool[n - i - 1]
```

Now lets rewrite this for `k = n`, so that we have one variable less to care about.

```python
n = len(x)
result = [None] * n  # <--
pool = list(x)
for i in range(n):  # <--
    j = randbelow(n - i)
    result[i] = pool[j]
    pool[j] = pool[n - i - 1]
```

Next, we redefine `i -> n - i - 1`. That also means that the interval `[0, n - 1)` now becomes `[n - 1, 0)`, or equivalently `(n, 1]`. So, `i` is now running from `n - 1` to `1`.

```python
n = len(x)
result = [None] * n
pool = list(x)
for i in reversed(range(1, n)):  # <-- i in [1, n), i.e. reversed (n, 1]
    j = randbelow(n - (n - i - 1))  # <--
    result[n - i - 1] = pool[j]  # <--
    pool[j] = pool[n - (n - i - 1) - 1]  # <--
```

which becomes


```python
n = len(x)
result = [None] * n
pool = list(x)
for i in reversed(range(1, n)):  # <-- i in [1, n), i.e. reversed (n, 1]
    j = randbelow(i + 1)  # <--
    result[n - i - 1] = pool[j]  # <--
    pool[j] = pool[i]  # <--
```

Now let's rearrange the code a bit:


```python
n = len(x)
result = [None] * n
pool = list(x)
for i in reversed(range(1, n)):
    j = randbelow(i + 1)
    result[n - i - 1], pool[j] = pool[j], pool[i]  # <--
```

The reasoning why `result[n - i - 1]` is equivalent to `result[i]` is, that we just write the elements into `result` in reverse. The order does not matter.

```python
n = len(x)
result = [None] * n
pool = list(x)
for i in reversed(range(1, n)):
    j = randbelow(i + 1)
    result[i], pool[j] = pool[j], pool[i]  # <--
```

Since we know by using `randbelow` that `0 <= j < i + 1`, we also know that the elements `result[i + 1]` and `pool[i + 1]` will never be touched again in the loop steps afterwards (we are going from `n` to 1), so we can also just use `x` instead of `result` and `pool`:

```python
n = len(x)
for i in reversed(range(1, n)):
    j = randbelow(i + 1)
    x[i], x[j] = x[j], x[i]  # <--
```

Et voilà, we have our definition of `rng.sample`, also [as stated above](https://github.com/ASPP/pelita/pull/856#issuecomment-2740654998).
The loops are doing the same thing, and since they do it for every step equivalently, it also holds when we assign `n --> k` with `k <= n` again.

Following my script, I wondered whether the sampled gaps are equal to the reverse of shuffled gaps when using the same seed:

```python
SEED = 1
rng = Random(SEED)

rng.shuffle(wall)
shuffled_gaps = wall[:ngaps]
sampled_gaps = rng.sample(wall, k=ngaps)

shuffled_gaps == sampled_gaps[::-1]
```

but it seems to make a difference whether the code runs through `[0, n - 1)` or `[n - 1, 0)`.
I suspect that the `randbelow` is the reason for that, since it is not symmetric to our index redefinition `i --> n - i - 1`. If we had an equivalent `randabove`, maybe it would work out.

But yeah @otizonaizit, I hope to have removed any of your doubts about rewriting the code from

```python
ngaps = max(1, ngaps)                                                        
wall_pos = list(range(max_length))                                           
rng.shuffle(wall_pos)                                                        
gaps_pos = wall_pos[:ngaps]                                                  
for gap in gaps_pos:                                                 
if vertical:                                                             
    wall.discard((pos, ymin+gap))                                        
else:                                                                    
    wall.discard((xmin+gap, pos))                                        
```

to 

```python
gaps = rng.sample(wall, k=max(1, ngaps))                                           
for gap in gaps:                                                           
     wall.remove(gap)
```

:grin:

otizonaizit:
@Debilski : we think this is ready to merge. Anything against it, @Debilski ?

Debilski:
Can you (or I?) rebase onto main before reviewing? So we can get rid of the three merge commits in the history?

Debilski:
Hmm, `pelita-createlayout` is broken. Can you fix this and ideally add a test?

otizonaizit:
> Hmm, `pelita-createlayout` is broken. Can you fix this and ideally add a test?

good catch. When #867 is merged `pelita-createlayout` is obsolete (just use `pelita --save-layout` instead). I'll get rid of `pelita-createlayout` in #867

otizonaizit:
> Can you (or I?) rebase onto main before reviewing? So we can get rid of the three merge commits in the history?

I get all sort of rebase conflicts if I do... not sure it is worth the trouble...

Debilski:
I only have one conflict in the first commit and that is easily fixed (#main contains the correct variant). I’ll push it here.

Debilski:
> > Hmm, `pelita-createlayout` is broken. Can you fix this and ideally add a test?
> 
> good catch. When #867 is merged `pelita-createlayout` is obsolete (just use `pelita --save-layout` instead). I'll get rid of `pelita-createlayout` in #867

I see. So, how do I test it? Or I’ll just believe you that it works? :)

You marked two tests as xfail. Can they be fixed or deleted?

otizonaizit:
Ah, right. I'll add a test for the `--save-layout` option in #867. The xfail tests were there for me to remember what needs to be tested in the new code. I'll take a look. I think I can do it only tomorrow afternoon. So in case you or Jakob want to have a go at it before that, feel free.

Debilski:
I fixed test_distribute_food. Not sure about the other one. `add_food` doesn’t exist anymore as a function.

jbdyn:
Thanks @Debilski for checking. If I am not mistaken, there is nothing left to do, right?

Debilski:
(Nothing against the TODO in general but I think they are outdated)

otizonaizit:
> Fixed a few spelling errors and did some minor linting.
> 
> One thing: There are still a bunch of TODO comments in the file as well as references to a find_chambers function that does not exist anymore. If you could take care of this I’ll merge it later.

Yep, I know. The TODOs should stay though. I am not sure I will have enough time to implement them in time for the release.

After the release I plan to refactor maze generation again, probably together with @jbdyn .

This will break backward compatibility with the numpy generated mazes, which we still have in this PR (same seed -> same maze). On the other hand, it will increase readability so that the poor guy that needs to change something 10 years down the line won't need 4 weeks of hard work to understand what is going on ;-)

I also plan to expand the test suite for maze generation. The TODOs are very detailed because of this. It's just temporary.

jbdyn:
> as well as references to a find_chambers function that does not exist anymore

The TODOs are now consistent with the code.

Debilski:
Ok, as long as you understand what these TODOs mean, it is ok. Because I think it isn’t clear from the file itself whatever that numpy compatibility is. :)

I’ll merge.

PR #857, opened 2025-03-10
Fix maze_generator still using old seed argument to get_new_maze

Debilski:
Plus bonus commit: rename `get_new_maze` → `create_maze` which to me follows more common API conventions. If this is merged, I suggest that #856 is rebased.

@otizonaizit Do you know why you close the stdout stream? I assume flushing is enough?


otizonaizit:
> @otizonaizit Do you know why you close the stdout stream? I assume flushing is enough?

Yes, flushing is enough.

PR #860, opened 2025-03-11
WIP: Standalone viewer with overlay graphics

### LINKED ISSUE #854: Add better Tk *viewer* mode

Debilski:
I hope to have a beta ready until tomorrow. The basic idea is that in one window, one starts a Tk viewer as:

    pelita-tkviewer --standalone-mode tcp://127.0.0.1:8888

which binds to port 8888 and will display any layout that is sent to it. This means that one should keep this window opened but can quickly send layout snapshots from for example an open ipython session.

In addition, I added a small overlay protocol that can (currently) colourise the background of the individual cells. Nothing fancy but we can now have a different colour for each chamber, for example.

image

The feature is currently Pelita-developer oriented. Not for users.

Closes #854 .

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -93,7 +93,7 @@ def controller_exit(state, await_action='play_step'):
 
 def run_game(team_specs, *, layout_dict, layout_name="", max_rounds=300,
              rng=None, allow_camping=False, error_limit=5, timeout_length=3,
-             viewers=None, viewer_options=None, store_output=False,
+             viewers=None, store_output=False,
              team_names=(None, None), team_infos=(None, None),
              allow_exceptions=False, print_result=True):
     """ Run a pelita match.
@@ -141,8 +141,6 @@ def run_game(team_specs, *, layout_dict, layout_name="", max_rounds=300,
     viewers : list[viewer1, viewer2, ...]
            List of viewers to attach to the game. Implemented viewers: 'ascii',
            'progress', tk'. If None, no viewer is attached.
-
-    viewer_options : do not use!
 
     store_output : False or str
                 if store_output is a string it will be interpreted as a path to a
@@ -201,7 +199,6 @@ def run_game(team_specs, *, layout_dict, layout_name="", max_rounds=300,
                        allow_camping=allow_camping,
                        error_limit=error_limit, timeout_length=timeout_length,
                        rng=rng, viewers=viewers,
-                       viewer_options=viewer_options,
                        store_output=store_output, team_names=team_names,
                        team_infos=team_infos,
                        print_result=print_result)
@@ -221,15 +218,8 @@ def run_game(team_specs, *, layout_dict, layout_name="", max_rounds=300,
     return state
 
 
-def setup_viewers(viewers=None, options=None, print_result=True):
+def setup_viewers(viewers, print_result=True):
     """ Returns a list of viewers from the given strings. """
-    if viewers is None:
-        viewers = []
-
-    if options is None:
-        options = {}
-
-    zmq_publisher = None
 
     viewer_state = {
         'viewers': [],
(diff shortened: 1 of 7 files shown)


Debilski:
Example session:

* Terminal 1:

```bash
# Starting a viewer in standalone mode
pelita-tkviewer --standalone-mode tcp://127.0.0.1:8888
```

* Terminal 2:

```bash
# We can now run a game on two screens at the same time
pelita --publish tcp://127.0.0.1:8888
```

```bash
# Or even on one screen and inside the terminal
pelita --publish tcp://127.0.0.1:8888 --ascii
```


Or finally, run this python script directly or in an interactive session and enjoy all the colourful chambers

```python
from random import randint
import time

import networkx as nx

from pelita.maze_generator import create_maze
from pelita.layout import parse_layout
from pelita.network import ZMQPublisher
from pelita.team import walls_to_graph


def find_chambers(G: nx.Graph, shape):
    w, h = shape

    main_chamber = set()
    chamber_tiles = set()

    for chamber in nx.biconnected_components(G):
        max_x = max(chamber, key=lambda n: n[0])[0]
        min_x = min(chamber, key=lambda n: n[0])[0]
        if min_x < w // 2 <= max_x:
          # only the main chamber covers both sides
          # our own mazes should only have one central chamber
          # but other configurations could have more than one
          main_chamber.update(chamber)
          continue
        else:
          chamber_tiles.update(set(chamber))

    # remove shared articulation points with the main chamber
    chamber_tiles -= main_chamber

    # combine connected subgraphs
    subgraphs = G.subgraph(chamber_tiles)
    chambers = list(nx.connected_components(subgraphs))

    return chambers, chamber_tiles

# Generate a new layout
maze = create_maze(height=16, width=32, nfood=30, dead_ends=True)
layout = parse_layout(maze)

# Find the chambers
graph = walls_to_graph(layout['walls'])
chambers, tiles = find_chambers(graph, layout['shape'])

# Colourize the cells in each chamber and generate an overlay
layout['overlays'] = [{'fill': '#%06X' % randint(0, 0xFFFFFF), 'pos': list(pos)} for pos in chambers]


# Send the layout with overlay to our viewer in Terminal 1
publisher = ZMQPublisher('tcp://127.0.0.1:8888', bind=False)
time.sleep(0.1) # small unfortunate race condition. The publisher must be alive for a bit or it will drop the message
# Once it is connected, it can however be reused
publisher.socket.send_json({'__action__': 'observe', '__data__': layout})

```

Debilski:
Doesn’t have the `--only-show-layout` and `--delay` options that were suggested in #854 yet but could be merged in principle.

Sending a layout to a viewer is a little unpolished to use but two lines of code is not too bad for development.

@otizonaizit @jbdyn let me know if this is still helpful for your layout-generation exercises

jbdyn:
@Debilski Looks very cool! Visualization was very helpful before, and with that we don't need to place food pellets for that.

Happy to see this being merged, thank you! :+1:

Debilski:
@otizonaizit any remarks here? Otherwise I’ll merge.

If necessary, we can also add more overlay styles but I’m not sure what would be useful without being too much

otizonaizit:
> @otizonaizit any remarks here? Otherwise I’ll merge.
> 
> If necessary, we can also add more overlay styles but I’m not sure what would be useful without being too much

LGTM, merge!

PR #867, opened 2025-04-24
Live maze generation

otizonaizit:
Use #856 in pelita.

This PR gets rid of the pre-computed layouts and introduces live maze generation. . We also got rid of all the functionality that are obsoleted by live maze generation, including all of the layout management routines.

The `layout_name` attribute of the game state is gone too.

New functionality:

  - A `--food` option to `pelita` CLI, to dynamically choose the number of "trapped" and total food pellets.
  - `--size` accepts a `<width>×<height>` specification in addition to the standaard `small`, `normal` and `big`.
  - the defaults for maze size and amount of food are now global variables in `game.py` instead of being implicitly defined by default values within functions.
  - The user can save the randomly generated layout to file if needed with `--save-layout`.

 Work done together with @jbdyn

To be merged after #856, as it contains that branch too.


otizonaizit:
The failing tests in [`pelita_template`](https://github.com/ASPP/pelita_template) are fixed in https://github.com/ASPP/pelita_template/pull/84

pelita/scripts/pelita_main.py
```diff
@@ -249,7 +260,7 @@ def long_help(s):
                            help='Stop when a bot has been killed.')
 
 viewer_settings = parser.add_argument_group('Viewer settings')
-viewer_settings.add_argument('--geometry', type=geometry_string, metavar='NxM',
+viewer_settings.add_argument('--geometry', type=w_h_string, metavar='NxM',
```
Debilski, line 263: Edge case: This allows `pelita --geometry small`

pelita/scripts/pelita_main.py
```diff
@@ -224,16 +233,18 @@ def long_help(s):
...
 
 layout_opt = game_settings.add_mutually_exclusive_group()
 layout_opt.add_argument('--layout', metavar='LAYOUT',
-                        help='Use maze layout specified in LAYOUT. LAYOUT can be'
-                             ' a file containing a valid layout or the name of a '
-                             'built-in layout. Use --list-layouts to get a list '
-                             'of all available layouts.')
-layout_opt.add_argument('--size', metavar='STRING', default='normal',
+                        help='Use maze layout specified in LAYOUT. LAYOUT is'
+                             ' a file containing a valid layout string.')
```
Debilski, line 244: I think we can simplify this to `'Use layout from FILE'` and have `'FILE'` (or FILENAME) as the metavar. I think it is obvious that the layout file contains something like a layout string.

pelita/scripts/pelita_main.py
```diff
@@ -208,14 +217,14 @@ def long_help(s):
...
                     metavar='FOLDER')
-parser.add_argument('--list-layouts', action='store_true',
-                    help='List all available built-in layouts.')
 parser.add_argument('--check-team', action="store_true",
                     help=long_help('Check that the team is valid (on first sight) and print its name.'))
 parser.add_argument('--append-blue', type=str, metavar='INFO', default=None,
                     help=long_help('Append info about the blue team (such as group id).'))
 parser.add_argument('--append-red', type=str, metavar='INFO', default=None,
                     help=long_help('Append info about the red team (such as group id).'))
+parser.add_argument('--save-layout', help=long_help('Save current layout to a file'),
```
Debilski, line 226: 60:40 feeling to prefer --store-layout or --write-layout since we need to give a file name. (--save feels more implicit to me like an automatic option that already includes the file name, but maybe that’s just me.)

pelita/scripts/pelita_main.py
Debilski, line 1: The part with ‘Layout specification’ can be removed from the `--help`

pelita/ui/tk_canvas.py
```diff
@@ -895,8 +892,6 @@ def draw_status_info(self, game_state):
                 else:
                     self.ui_bot_traffic_lights_canvas.itemconfig(circle, fill="#bbb")
 
-        layout_name = "–" if game_state.get("layout_name") is None else game_state.get("layout_name")
```
Debilski, line 898: I think we might add the seed now to the UI? (Possibly together with the food specification.) We should make it possible to recreate a maze from a screenshot, if only for dealing with bug reports. I don’t think it needs to be as prominent as the layout name, though.

Debilski (commented):
Few comments to the docs, looks good otherwise. We should rebase once #856 has been merged.

Do we know how it performs on a slower computer?

pelita/scripts/pelita_main.py
otizonaizit: fixed in 3efd0dcc

pelita/scripts/pelita_main.py
otizonaizit: fixed in bbb10cca

pelita/scripts/pelita_main.py
otizonaizit: yes... I think it's ok to live with this? You can't always prevent people from shooting themselves in the foot...

pelita/scripts/pelita_main.py
otizonaizit: fixed in eb6f47cf

pelita/ui/tk_canvas.py
otizonaizit: We tried it. We convinced ourselves that it was not useful. To be useful in anyway the text from the GUI would have to be selectable so that you can copy it with ctrl+c, which is not. Typing the seed from a screenshot is not going to be nice either. I don't think any of this is a problem. The seed is printed very prominently in the CLI, and you never use pelita without the CLI.
I am not sure it is worth it to try to make the GUI a sort of `repr` of the game. The CLI is good enough for this purpose in my opinion.

otizonaizit:
Regarding adding a test for `--store-layout`, a.k.a. `--save-layout` as requested in https://github.com/ASPP/pelita/pull/856#issuecomment-2828231058 I decided to give up. We do not have the test infrastructure to test the CLI at the moment. There is `call_pelita` in `tournament.py`, but that function saves a different purpose and does not allow to pass arbitrary option to the `pelita` main script.
If we migrate to `click`, which would make me very very happy, then testing the CLI is going to be easy.

otizonaizit:
> Few comments to the docs, looks good otherwise. We should rebase once #856 has been merged.
Could you take care of the rebase?

> Do we know how it performs on a slower computer?
Not really, but on my computer I tested start-up times (running under `pyinstrument` with stopping bots and `--rounds 1`). Of the total 0.5 seconds, the maze generation took 0.002 seconds. We don't need to worry about this.
Interestingly, half of the time is wasted in importing `networkx` and the other half in setting up the network connection with `zmq`.

pelita/ui/tk_canvas.py
Debilski: Right, makes sense. Although on macOS any text is now selectable in a screenshot, so it is not too much of a problem.

If we find that we need it, we may still add it later.

Debilski:
> Regarding adding a test for `--store-layout`, a.k.a. `--save-layout` as requested in [#856 (comment)](https://github.com/ASPP/pelita/pull/856#issuecomment-2828231058) I decided to give up. We do not have the test infrastructure to test the CLI at the moment. There is `call_pelita` in `tournament.py`, but that function saves a different purpose and does not allow to pass arbitrary option to the `pelita` main script. If we migrate to `click`, which would make me very very happy, then testing the CLI is going to be easy.

Not sure about that. To have it working with click, you will need to write a separate function to handle the CLI options (or, in our case of a rather convoluted CLI: a few different functions), rather than putting everything inside one big `main` function, which obviously makes testing simpler. Of course, we could already do that now, if we wanted to. :)

(I kind of plan to move away from click again where I introduced it in pelita_player and pelita_server, because it feels messy with our variadic CLI and you can’t really reuse the annotated functions, so everything has to be written twice basically.)

Debilski:
I adjusted the message with the seed a little.

```
❯ pelita --null
Replay this game with --seed 7811846223299860753
ᗧ blue team: 'Smart Eating Players'
ᗧ red team: 'Smart Random Players'
Finished after 222 rounds. 'Smart Random Players' won over 'Smart Eating Players'. (45:30)
```

```
❯ pelita --store-layout /tmp/x
Regenerate this layout with --seed 7269427535039484812
```

When a seed is given, nothing is printed. Feels a bit empty but I don’t really know what to expect here.

```
❯ pelita --seed 123 --store-layout /tmp/x
```

Since we currently don’t forbid it, it is possible to create a layout from a layout. This doesn’t use the seed, so we don’t print it here either. (We could obviously forbid it in argparse, but it may be useful for normalising hand-written layouts in that it removes superfluous space.) [*]

```
❯ pelita --layout /tmp/x --store-layout /tmp/x
```

Should we make it more verbose and tell what we are doing? (Maybe something for a `--verbose` flag; might be useful for testing to see, eg. how long it took to generate a layout and how many chambers/what food distribution it has.)

If you’re ok with it, I’d also add `--store-layout -` as a shortcut for writing to stdout.


[*] Thinking about this: Do we want to encourage overwriting any filename that is passed to us? Or should we better play it safe and only write new files.

otizonaizit:
> I adjusted the message with the seed a little.

That's a nice improvement!

> When a seed is given, nothing is printed. Feels a bit empty but I don’t really know what to expect here.

I think it's OK if it's empty.



> Since we currently don’t forbid it, it is possible to create a layout from a layout. This doesn’t use the seed, so we don’t print it here either. (We could obviously forbid it in argparse, but it may be useful for normalising hand-written layouts in that it removes superfluous space.) [*]
> 
> ```
> ❯ pelita --layout /tmp/x --store-layout /tmp/x
> ```

I think it's OK. That's really an edge case and as you say, this is practically a no-op, so no surprises.

> Should we make it more verbose and tell what we are doing? (Maybe something for a `--verbose` flag; might be useful for testing to see, eg. how long it took to generate a layout and how many chambers/what food distribution it has.)

I think that for this sort of things you'll have to write a custom script. No need to mess with the pelita CLI. After all, `import pelita; pelita.maze_generatort.generate_maze()` is not too hard to write, and after that you can do all the analysis and/or profiling you want.

> If you’re ok with it, I’d also add `--store-layout -` as a shortcut for writing to stdout.

Sure, that's also a good addition. Actually you could make it such that if no argument is passed, then it prints to stdout. And also support `-` for POSIX compliance.
 
> [*] Thinking about this: Do we want to encourage overwriting any filename that is passed to us? Or should we better play it safe and only write new files.

I think we should just overwrite. If I am playing around with different sizes and/or food for testing I may re-use the same filename over and over again. Consenting adults and stuff... if you want to babysit the user you can issue a warning, but that doesn't really help if the file is overwritten already ;-)

Debilski:
> > Should we make it more verbose and tell what we are doing? (Maybe something for a `--verbose` flag; might be useful for testing to see, eg. how long it took to generate a layout and how many chambers/what food distribution it has.)
> 
> I think that for this sort of things you'll have to write a custom script. No need to mess with the pelita CLI. After all, `import pelita; pelita.maze_generatort.generate_maze()` is not too hard to write, and after that you can do all the analysis and/or profiling you want.

Yeah, ok.

> > If you’re ok with it, I’d also add `--store-layout -` as a shortcut for writing to stdout.
> 
> Sure, that's also a good addition. Actually you could make it such that if no argument is passed, then it prints to stdout. And also support `-` for POSIX compliance.

Support for - has been added.

Without an argument, the name `--store-layout` feels wrong again though, so I didn’t set any default there. :)

> > [*] Thinking about this: Do we want to encourage overwriting any filename that is passed to us? Or should we better play it safe and only write new files.
> 
> I think we should just overwrite. If I am playing around with different sizes and/or food for testing I may re-use the same filename over and over again. Consenting adults and stuff... if you want to babysit the user you can issue a warning, but that doesn't really help if the file is overwritten already ;-)

`rm layout ; pelita --store-layout layout` is also not too hard to write. ;)

Almost good to go, we have one final thing to solve though:

```
def test_maze_generation_roundtrip():
    maze = mg.generate_maze()
    maze_str = pl.layout_as_str(**maze)
    maze_from_str = pl.parse_layout(maze_str)

    assert maze == maze_from_str
```

Should work but fails because `generate_maze` returns a tuple of tuples for walls, whereas `parse_layout` returns a set or tuples. For the internal stuff, a set makes more sense (and we wouldn’t need 9e9e197 for example). What should we do?

Debilski:
Ah wait. It works. But the docs to `parse_layout` are wrong. Ok, then.

Still, I am kind of unhappy with the amount of non-sets that we use internally. :)

otizonaizit:
ready to merge?

PR #869, opened 2025-04-24
Make grid labels visible to older people

otizonaizit:


otizonaizit:
the failing test is unrelated. Unfortunately one of the tests depending on randomness that sometimes fail. There must be a better way for doing this. I just run the same test 1000 times on my machine and it didn't fail once.

Debilski:
Open an issue about the failing test and just click on rerun in CI.

That said: This label size cuts off the two digit numbers for me. (Does it look fine for you?) We may have to fiddle with the margins.

otizonaizit:
> Open an issue about the failing test and just click on rerun in CI.
> 
> That said: This label size cuts off the two digit numbers for me. (Does it look fine for you?) We may have to fiddle with the margins.

ouch, that's annoying...

otizonaizit:
maybe it is enough to add some padding to the canvas? I have no idea where to do it in the code though...

Debilski:
Well, I’d say your `label_size = 7` is cut off as well. Solution lies in changing `MeshGraph.padding`

Debilski:
I took the liberty to change the margin/padding of the header because I think it looks a bit too far away with the new canvas padding when the grid is off. Feel free to keep or drop it. :) Otherwise good to go. We should think about retuning a few things about the UI though.

Debilski:
Before:
image

After:

image
image

PR #870, opened 2025-04-25
add `Bot.has_exact_position` attribute

otizonaizit:
The `Bot.is_noisy` attribute has historically being perceived as confusing. The attribute `Bot.has_exact_position`, internally implemented simply as `not bot.is_noisy` conveys the intent in a clearer way.

More discussion about this in this issue in the pelita_template repo:

https://github.com/ASPP/pelita_template/issues/85

### DIFF
--- a/pelita/team.py
+++ b/pelita/team.py
@@ -567,6 +567,7 @@ class Bot:
         self.team_time = team_time
         self.error_count = error_count
         self.is_noisy = is_noisy
+        self.has_exact_position = not is_noisy
         self.graph = graph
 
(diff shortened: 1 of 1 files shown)


Debilski:
Language-related discussion in https://github.com/ASPP/pelita_template/issues/85 . I think people are using the term and variable ‘is close’ already when they are checking for imminent danger nearby, so we shouldn’t overload the phrase too much. As much as I find noisiness confusing, it is the one time we that we found a unique word for a concept in Pelita. :)

otizonaizit:
ready to merge?

Issue #832, opened 2024-10-05
Get rid of non-fatal errors


otizonaizit:
At the moment a bot can generate different kind of "errors":

1. it returns an illegal position
2. it gets a timeout
3. it raises an Exception

The code (and the rules) are complicated by the fact that we treat these errors differently. Errors of type 1 and 2 result in a random move and are counted. After 5 errors of this type the game is over. Errors of type 3 result in an immediate game over.

After many years seeing bots implemented by students, I think that it would be more useful to handle errors of type 1 as fatal like type 3: immediate game over. This helps in debugging, because you get immediate feedback when you return an illegal position, instead of having to fish out the warning printed by pelita among the other thousand lines printed by your bot in the development phase. Errors of type 1 are easy to avoid (check `next_pos in bot.legal_positions`). This I think is an uncontroversial change that would simplify pelita code **and** the rules that we need to explain quite a bit.

Regarding errors of type 2 I think that they also should be made fatal. It almost never happens that bots generate timeouts, except the network bots – but this is a different issue.

If all errors would be made fatal we would have easier rules to explain, less clutter in the UI (because we don't have to show the error counts) and help students to catch errors earlier so that they write more robust bots.

Let's start a discussion!

Debilski:
Open to making wrong moves fatal, but I think one reason we have it is because of the timeout handling. The idea being that after a timeout the bot will still attempt to send the move and then there is a small danger that the reply gets mixed up with the next, regular reply from the bot. We just play it safe here. This should not happen but I’m not sure if we have tests for it.


I’m more strongly against making timeouts fatal. If you have a bot that does a lot of analyses and caching in the first round, then the time that it takes can be very much dependent on the layout. You can hit 2.5 seconds regularly but then in the tournament you get the one maze that takes 3.2 seconds.

Pretty sure we would have seen spurious timeouts in the last tournament, had we run it on one of the students laptops and not on mine, which happens to be two to three times faster.

otizonaizit:
And what if we make the timeouts never fatal, i.e. we stop counting them?

I am aiming at a simplification of the rules **and** of the code. Also, in case of timeout, instead of a random move, we could just let the bot keep its position. The fact that we execute a random move is one of the last few things that make games (and the tournament, see #833 ) strictly non-reproducible. A timeout can be due to the host being busy with other things.

So, my proposal: we make illegal positions fatal and allow timeouts. In case a bot gets a timeout, it simply stays where it is. No fatal error after N timeouts.

[...]

PR #871, opened 2025-04-26
Make returning an illegal position a fatal error

otizonaizit:
This is the minimal implementation of making illegal moves a fatal error.

Given that the only other possible non-fatal error are timeouts, it may be worth it to refactor the error handling. I decided not to try this in this PR, because it would be a lot of work and I'd like this minimal implementation to go into the next release, so that we can start seeing the benefits right away. After all the error handling is not a user-facing feature.

In the process I fixed some tests that were still using the old API for the move function and were working because of the side effect of having `error_limit=5`, but these bots were returning invalid positions all along...

If and when this PR is merged I'll have a go at updating the docs in pelita_template.

Fixes: #832

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -764,7 +764,7 @@ def play_turn(game_state, allow_exceptions=False):
         game_state = apply_move(game_state, position)
 
         # If there was no error, we claim a success in requested_moves
-        if (round, turn) not in game_state["errors"][team]:
+        if (round, turn) not in game_state["errors"][team] and not game_state['fatal_errors'][team]:
             game_state['requested_moves'][turn]['success'] = True
 
         # Check again, if we had errors or if this was the last move of the game (final round or food eaten)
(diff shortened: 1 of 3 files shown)


test/test_game.py
```diff
@@ -335,7 +335,7 @@ def test_multiple_enemies_killing():
     ########
     """
     # dummy bots
-    stopping = lambda bot, s: (bot.position, s)
+    stopping = lambda bot, s: bot.position
```
Debilski, line 322: The stopping players ‘worked’ because they were only used as dummies to get through `setup_game`. The game was then simulated with `apply_move` without ever calling the function. I suppose we could have used a more obvious dummy for this.

pelita/game.py
```diff
@@ -837,27 +837,22 @@ def apply_move(gamestate, bot_position):
...
+                    'round': n_round,
+            }
+            gamestate['fatal_errors'][team].append(exception_event)
 
     # only execute move if errors not exceeded
     gamestate.update(check_gameover(gamestate))
     if gamestate['gameover']:
         return gamestate
 
-    # Now check if we must make a random move
```
Debilski, line 853: You need this section to handle slow bots. Without it the whole game crashes on a timeout.

otizonaizit: yep, fixed

otizonaizit: I forced pushed to recover the code that handled random moves in the face of non-fatal errors. I also re-wrote the corresponding test to use timeouts instead of illegal moves to test that the error limit still works

test/test_game.py
otizonaizit: No, no, `apply_move` was called in some tests and the check `if bot_position not in legal_moves` was failing, because `(position, {})` could not be found in the list of tuples `legal_positions`. This was filing an error for the team. The simulated games only had one round, though, so the error never became fatal and all other checks in the tests were still correct.
they started failing for me when I transformed the check `if bot_position not in legal_moves` into a fatal thing. Then the tests couldn't proceed and would fail early.

Debilski: Yes, in some tests it failed. But in most others it didn’t. Doesn’t really matter but we can refactor the tests anyway at some point.

otizonaizit:
Ready to merge?

Debilski:
Curious how this will break things. :)

PR #874, opened 2025-04-28
Linting, project cleanup

Debilski:
Work in progress. (Not a draft so that I can have CI.)


Debilski:
Ruff fails because of #873 (our failure to detect this error is the reason I am adding this here :) )

otizonaizit:
Hey, I am not so sure this is a good idea. Having a big commit that reformats and refactors all over the place in the project makes it hard to use `git blame` and in general to make sense of changes down the line. Also, it blocks development until the big reformat is done.

Why not taking a more gentle approach and just reformat/refactor a file when you are touching that file for a reason, for example to implement a new feature? This way the improvement is incremental, and changes are localized and easier to make sense of.

It is good that ruff caught that bug in `run_background_game`, but I still think the costs that this approach brings with it are quite high and don't compensate the discovery of the occasional bug.

Regarding the kind of linting, we may have to have a discussion on which formatting we want. I have the tendency to try and adapt to the formatting a file is already written in. If we want to move to a policy where `ruff` is king, we need to decide it. But then maybe it would be better to have a git-hook to do the formatting. Otherwise this rises the bar even more for new contributors.

It is the same discussion as the one about adding types. There are pro and contra.

Debilski:
Eh, but I’m not reformatting. This is 80% dead code removal and running isort (which just fits well with removing unused imports).

The point is that these changes make `ruff check` already succeed, because our code isn’t too bad, so we might as well use it in CI. (Minus long lines and ambiguous variables, which I think we can rightfully ignore.)

The one controversial change is this https://docs.astral.sh/ruff/rules/multiple-statements-on-one-line-colon/ I don’t really care about it either way, so I might as well put it on the ignore list and revert the changes that I made.


> Why not taking a more gentle approach and just reformat/refactor a file when you are touching that file for a reason, for example to implement a new feature? This way the improvement is incremental, and changes are localized and easier to make sense of.

But that’s neither happening nor is it what we want. You could have removed 50 unused lines from tk_canvas when you got rid of the layout label but I don’t think it this would have been a better time to do it.

Now the next time we are messing around in some UI code, we get a nice reminder that there is some left-over code that needs some cleaning, and indeed, we will then have the chance to add it to the PR to which these changes belong.

> But then maybe it would be better to have a git-hook to do the formatting. Otherwise this rises the bar even more for new contributors.

Sure, you explain the git-hook to new contributors then … ;)

If a new contributor adds dead code, we wouldn’t accept it now either. I really fail to see the impact here.

Debilski:
Good to go from my side. I fixed some spelling mistakes, cleaned up dead code and reformatted the imports.

`MANIFEST.in` is gone
`pyproject.toml` contains entries for doc and test dependencies

CI uses `ruff check` to warn us about mistakes in the code. If we are unhappy about a warning, we can either decide to put in on a global ignore list (currently long lines and ambiguous variables are ignored) or they can be disabled on a per-line basis with, eg. `# noqa: F821`. I find that ruff tends to complain only about code that already looks bad to the eye, so I do not think it is a big deal to have it running in CI.

Debilski:
Remark: When removing dead code (unused variables), I chose one of three options:
a) deleting the line
b) prefixing the variable with an underscore (in cases when it is illustrative that a function has a return value)
c) commenting out the line but leaving it in the code (when it might be useful later)

Issue #881, opened 2025-05-22
Rendering problems on (some) MacOS clients


otizonaizit:
On some MacOS clients the text on top of the maze with errors, times, Exceptions, etc. is not rendered...

On some other MacOS clients it works though, and I couldn't find a pattern to predict if it would work or not...

@Debilski : any idea how to debug this?

Examples:

![Image](https://github.com/user-attachments/assets/3e30aec2-2d2f-4e9b-bbdd-c395525a80c6)

![Image](https://github.com/user-attachments/assets/dc848643-9397-470f-9f4c-9377d008098a)

Debilski:
If we remove all background=white and fill=black from the tk viewer, we get this experience on macOS when switching to dark mode.

https://github.com/user-attachments/assets/67097a19-fbde-48c2-a588-5bcb929037ac

Also somewhat workable. Most notably, however, the default background color is not #FFFFFF white but something more greyish. If we override this we will have to do more work to handle the transition to dark mode.

otizonaizit:
hmm, if you think that the dark mode is readable enough, it seems that this change will make the rendering on MacOS so much better even without dark mode, right?

Is this affecting Windows and Linux rendering though? If not, I think we could go for it, even if the walls are barely readable in the video, maybe they need a white stroke paint (a thin white border I mean): if the color is specified explicitly it won't affect the light mode, right?

Sorry, I don't have a Mac client to test.

Debilski:
It also has the grey hue on Linux. It think it looks a little too sad (even worse on Linux). It is also not really bug-free on macOS. the buttons have a whiter colour when the window goes to the background and becomes active again.

Maybe we should build a smaller UI prototype (just some button, label and a few colours) and then tweak it on a few platforms. There doesn’t seem to be an official way to detect dark mode. One workaround that I’ve read would involve keeping one element deliberately unstyled (such as the status font colour) and then, depending on whether this is a bright or a dark colour, adjust all (or most) other colours accordingly.

> it seems that this change will make the rendering on MacOS so much better even without dark mode, right?

I think a part of it can be fixed by also setting a background=white colour in the frame behind the buttons.

Debilski:
We could also move away from UI button elements and kind of draw the outlines ourselves. That way, we would get the most similar look on all OSs.

otizonaizit:
I would be for the minimal intervention that gets us to a place where darkmode in MacOS still allows to read the text without making the experience on Windows and Linux worse.
[...]

PR #882, opened 2025-05-22
Fix dark mode colours

Debilski:
Suggestion for #881 

Makes the buttons on macOS all in all nicer by removing the colouring completely, thereby getting rid of grey or black background mismatches. (There are alternative button sets like https://pypi.org/project/tkmacosx/ that solve the whole issue much nicer but it is not really clear, how stable these solutions are in the long run.)

As the default background colour for Tk has this sad grey, this obviously sets the whole status section apart. Can we live with that? Is this all in all worse?

Tk unfortunately doesn’t have any border-top or margin-top stylings to improve the appearance here. I had to resort to a dummy row on top of the status section with ipady=2 to add a few more pixels on top of the status bar, otherwise it wouldn’t really look ok.

image
image

Haven’t tested on Linux yet or with anything other than macOS.

### DIFF
--- a/pelita/ui/tk_canvas.py
+++ b/pelita/ui/tk_canvas.py
@@ -841,13 +841,13 @@ class TkApplication:
         padding = 15
 
         # middle colon
-        self.ui_header_canvas.create_text(center, top, text=":", font=(self._default_font, font_size), tags="title", anchor=tkinter.CENTER)
+        self.ui_header_canvas.create_text(center, top, text=":", font=(self._default_font, font_size), fill="black", tags="title", anchor=tkinter.CENTER)
 
         self.ui_header_canvas.create_text(center, top, text=left_team, font=(self._default_font, font_size), fill=BLUE, tags="title", anchor=tkinter.E)
         self.ui_header_canvas.create_text(center+2, top, text=right_team, font=(self._default_font, font_size), fill=RED, tags="title", anchor=tkinter.W)
 
-        self.ui_header_canvas.create_text(0 + padding, 20 + font_size, text=" " + left_status, font=(self._default_font, status_font_size), tags="title", anchor=tkinter.W)
-        self.ui_header_canvas.create_text(self.ui_header_canvas.winfo_width() - padding, 20 + font_size, text=right_status + " ", font=(self._default_font, status_font_size), tags="title", anchor=tkinter.E)
+        self.ui_header_canvas.create_text(0 + padding, 20 + font_size, text=" " + left_status, font=(self._default_font, status_font_size), fill="black", tags="title", anchor=tkinter.W)
+        self.ui_header_canvas.create_text(self.ui_header_canvas.winfo_width() - padding, 20 + font_size, text=right_status + " ", font=(self._default_font, status_font_size), fill="black", tags="title", anchor=tkinter.E)
 
     def draw_status_info(self, game_state):
         if "round" in game_state:
--- a/pelita/ui/tk_canvas.py
+++ b/pelita/ui/tk_canvas.py
@@ -209,40 +209,41 @@ class TkApplication:
         self.ui_sub_header.config(background="white")
 
         self.ui_status_canvas = tkinter.Frame(window, height=25)
-        self.ui_status_canvas.config(background="white")
 
         self.ui_game_canvas = tkinter.Canvas(window)
         self.ui_game_canvas.config(background="white", bd=0, highlightthickness=0, relief='ridge')
         self.ui_game_canvas.bind('<Configure>', lambda e: window.after_idle(self.update))
         self.ui_game_canvas.bind('<Button-1>', self.on_click)
 
-        self.ui_status_00 = tkinter.Frame(self.ui_status_canvas, background="white", padx=5, pady=0)
-        self.ui_status_01 = tkinter.Frame(self.ui_status_canvas, background="white", padx=5, pady=0)
-        self.ui_status_02 = tkinter.Frame(self.ui_status_canvas, background="white", padx=5, pady=0)
-        self.ui_status_10 = tkinter.Frame(self.ui_status_canvas, background="white", padx=5, pady=0)
-        self.ui_status_11 = tkinter.Frame(self.ui_status_canvas, background="white", padx=5, pady=0)
-        self.ui_status_12 = tkinter.Frame(self.ui_status_canvas, background="white", padx=5, pady=0)
-        self.ui_status_20 = tkinter.Frame(self.ui_status_canvas, background="white", padx=5, pady=0)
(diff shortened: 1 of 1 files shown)


Debilski:
… and in the first commit is the simple fix that makes the text for the error count always black.

otizonaizit:
If this doesn't break Linux or Windows I think we can live with it. This will make the status bar grey in Linux and Windows too, right? This is a downgrade, but I think it's less important. 

If you come up with a better solution for the buttons then we can revert it.

Or maybe just fix the black text issue in one separate PR, so we can bikeshed the buttons and status bar stuff a bit longer independent of the fix?

A point release with the fix would be welcome, so that my students with Macs can work properly, is it doable?

Debilski:
Yeah, will publish the first commit with the black text separately. (You can also tell the students to disable dark mode. :) )

I experimented with Labels instead of Buttons in Tk but they have a bit of a sluggish feel to them (if we want to have some tricks like changing the background color on hover). If I have some more ideas over the weekend, I’ll add the screenshots here and we can decide. Otherwise we will just release the first commit and do the rest later.

Debilski:
Maybe we can make the dark status bar work when we make it smaller? Like putting all buttons in one row?

otizonaizit:
No, let's go for grey everywhere then. Conditionals are evil ;)

Debilski:
Two small additional technical changes made:

The title section is not a separate canvas anymore; we draw the team names directly on the game canvas. This allows for tighter spacing (MeshGraph takes an offset parameter now).

Default geometry has been changed to make the cells square again.

otizonaizit:
Some issues on Linux still:

  - with `debug` on, the axis labels are almost overlapping with the status line...:

 
screenshot1

  - The `slower`, `faster` and `debug` button labels are in a smaller font than before, maybe too small? (see screenshot below)
  - The `quit` button has become smaller too, but that's the button people try to reach quite often, why make it smaller? (see screenshot below)

out

  - The grey is really ugly, but I guess it's just a matter of getting used to it?
  - The buttons on macOS are nicely spaced, the one on Linux are crammed together: I know it has been like this forever, but can't they be spaced on Linux too?
  - Given how platform dependent all this seems to be, we have to make sure that the appearance on Windows is also OK

otizonaizit:
This is for another time, but what about changing the button labels like this:
  - `PLAY/PAUSE` ➔ ⏯
  - `STEP` ➔ ⏭
  - `ROUND` ➔ get rid of this
  - `slower` and `faster`: we can find symbols for these two, maybe: « and » ?
  - `debug` could be moved on the right, near to the debug info?

Debilski:
> This is for another time, but what about changing the button labels like this:
> 
> * `PLAY/PAUSE` ➔ ⏯
> * `STEP` ➔ ⏭

This sign clearly means play until the end without stopping.

> * `ROUND` ➔ get rid of this
> * `slower` and `faster`: we can find symbols for these two, maybe: « and » ?

The canonical icons are 🐢 and 🐇.

> * `debug` could be moved on the right, near to the debug info?

I think slower and faster are also debug-mode buttons. (In the sense that in a normal game you should never need to press them.)

We can test it but I’m a little doubtful that this will have an improved look with Tk. The only thing that makes these oversized buttons work is that we have oversized text in them. :)

otizonaizit:
> > Some issues on Linux still:
> > 
> > * with `debug` on, the axis labels are almost overlapping with the status line...:
> 
> Ah, too bad. We are off by two rows of pixels compared to macOS here (macOS is still rather close but I think it works and doesn’t overlap). Will see if I can solve this with better font sizes.

yes, but please do not make the labels smaller: old people (me) still need to see it ;)

> > * The `slower`, `faster` and `debug` button labels are in a smaller font than before, maybe too small? (see screenshot below)
> 
> Yeah, I wanted to remove the visual weight from the second row there. (The ‘nothing selected’ text is also in a smaller font.) If it’s not readable (or you think it looks worse), we can revert it.

it's ok for the slower and faster buttons. The `debug` button is different though, because that's a button people use very very often and sometimes in a hurry.

> > * The `quit` button has become smaller too, but that's the button people try to reach quite often, why make it smaller? (see screenshot below)
> 
> I wanted to give the buttons to the left more space when the window is made smaller and adjusted the weights for the grid layout. (Resizing still doesn’t work too well, I haven’t figured out a good way to make all the buttons shrink at the same time.)

I don't know. To me it looked better before, but I am not terribly attached to that either.

> People who regularly use the quit button and still have trouble finding it, should really learn about the letter q. ;)

Yes. Is there anything bad to expect if we just remove the `QUIT` button? If closing the window has the same effect, I think we can even get rid of it completely.

> > * The grey is really ugly, but I guess it's just a matter of getting used to it?
> 
> ```
> FRAME_STYLE = {} if platform.is_macos else { "bg": "white" }
> 
> tk.Frame(window, *FRAME_STYLE)
> ```
OK, go for it!

> > * The buttons on macOS are nicely spaced, the one on Linux are crammed together: I know it has been like this forever, but can't they be spaced on Linux too?
> 
> In principle yes. The problem is: This will make them even further apart on macOS and if you’re not careful and use the wrong button styles (like height), they will change their look on macOS completely. It’s all a big mess.

Then maybe until we have the right™ solution we just go for the conditional above even in this case?

> > * Given how platform dependent all this seems to be, we have to make sure that the appearance on Windows is also OK
> 
> Windows and Linux looked similar but I am not sure how much this is actually dependent on how Python itself is installed.
> 
> > Another glitch in Linux: the button bar now has almost no padding anymore on the border with the maze. It was a bit more spaced before. Or maybe it wasn't, but because it was the same color, it wasn't noticeable. Now that it is grey, it seems a bit crammed to me...
> 
> Yeah, I agree. It kind of works on macOS I think but the effect on Linux is worse.

so, another conditional?

Debilski:
> > > * `debug` could be moved on the right, near to the debug info?
> > 
> > 
> > I think slower and faster are also debug-mode buttons. (In the sense that in a normal game you should never need to press them.)
> 
> no, really, I see a lot of students pressing faster faster faster when they test their bot without using the debug mode. They should learn `--fps`, yes, but pressing the button is easier for most ;)

No, I agree, people use it a lot and it should be easily reachable. But what I meant is, once you press it, you are – by my definition – not in a normal game anymore but you are debugging. Therefore the button location should be adjacent to the other debug tools and not next to the regular game controls.

That said, I think these buttons gives an illusion of power that is not really there, suggesting infinite steps of speed that are neither useful nor possible. Something like this would be more fitting:

image

OTOH, buttons are easier to press, so we should keep them I guess, but maybe reduce the number of available steps?

Debilski:
(Not tested on Windows yet)

Debilski:
image

Buttons on Linux are bigger (there is commented-out code for this in the PR but in the end I think this has been enough fine-tuning already)

Debilski:
Merge and make a point release then?

PR #889, opened 2025-06-06
Threaded CI engine

Debilski:
Work in progress, but gets the job done already (needs better data structures for result passing and the counting may still be broken).

Run 100 matches (10 at the same time):

    contrib/ci_engine.py run -n 100 -t 10

and then view the result:

    contrib/ci_engine.py print-scores

As it turns out, the old approach of simply running a bunch of ci_engine processes in parallel on a server is not really accurate (sqlite gets overwritten unless it is explicitly locked, in which case it will just crash). We basically need one and only one thread to do all the db handling.

TODO:

- [x] Better counting (new match selection)
- [x] Elo is broken
- [x] Hashing should be done concurrently
- [x] Breaks when not hashed
- [x] No detection when players have changed

### DIFF
--- a/.github/workflows/test_pytest.yml
+++ b/.github/workflows/test_pytest.yml
@@ -129,5 +129,5 @@ jobs:
     - name: Run ci_engine session
       run: |
         cd contrib
-        python ci_engine.py -n 5
+        python ci_engine.py run -n 5
       timeout-minutes: 5
--- a/contrib/ci_engine.py
+++ b/contrib/ci_engine.py
@@ -42,7 +42,9 @@ stabilized.
 """
 
 import argparse
+import collections
 import configparser
+import heapq
 import itertools
 import json
 import logging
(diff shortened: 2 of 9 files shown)


Debilski:
Code is a bit of a mess in some parts; there are multiple ways that concurrency is done and the whole database stuff feels very low level but it kind of works and allows one to quickly run 1000s of matches (given enough cores).

As a future task, I think we should give the `ci_engine` its own `pyproject.toml` so that we can maybe add SQLAlchemy or similar as a dependency (with the bonus of being able to use something other than sqlite as storage [*]) and make the database handling a bit nicer on the eyes and in general separate looping logic from output.

Debilski:
@otizonaizit Do you *want* to review this or should we just merge as it does hardly affect anything in Pelita main?

otizonaizit (approved):
OK, I just checked the changes in pelita main. They look fine.
You can merge if you are done with rebasing.

PR #894, opened 2025-06-11
Make maze generation work for small sizes

### LINKED ISSUE #893: maze generation fails if maze height < 6
```bash
$ pelita --size 10x5 --food 0:1 0 0
...
ValueError: max() iterable argument is empty
```

fix is on its way

otizonaizit:
Fixes #893

### DIFF
--- a/test/test_maze_generation.py
+++ b/test/test_maze_generation.py
@@ -249,20 +249,37 @@ def test_generate_maze(iteration):
     l2 = mg.generate_maze(trapped_food, total_food, width, height, rng=seed)
     assert l1 == l2
 
-@pytest.mark.parametrize('iteration', range(100))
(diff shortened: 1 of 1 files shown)


Debilski:
LGTM although the mazes of height 5 that are created don’t seem to have a lot of walls anyway.

otizonaizit:
> LGTM although the mazes of height 5 that are created don’t seem to have a lot of walls anyway.

Yes, given that we stop at partitions of size 3, and given that for height 5 minus external border minus internal border we are only left with 3 tiles, it is expected that you can only have walls on the internal border.

The whole thing is only happening in the edge case of height < 6, so it is almost never relevant.
But at least now it works also for that case.

PR #896, opened 2025-06-11
Restrict `bot.say` argument to a short ASCII alphanumeric string

### LINKED ISSUE #892: a specially crafted `bot.say` can make the GUI unreadable
With this bot we can destroy the GUI:

`color_bot.py`:
```python
TEAM_NAME = "Color"                                                              
                                                                                 
def move(bot, state):                                                            
    bot.say(('█'*500 + '\n')*500)                                                
    return bot.position                                                          
```

And this happens at the first turn:

Image

I think we need to impose restrictions on `bot.say` as we do for `TEAM_NAME`.

otizonaizit:
Fixes #892

### DIFF
--- a/pelita/team.py
+++ b/pelita/team.py
@@ -90,6 +90,36 @@ def walls_to_graph(walls, shape=None):
     return graph
 
 
+def sanitize_say(string):
+    """Make input string sane (for a certain definition of sane)"""
+    sane = []
+    # first of all, verify that the whole thing is valid unicode
+    # this should always be True, but who knows where do they get
+    # their strings from
+    try:
+        string.encode('utf8')
+    except UnicodeEncodeError:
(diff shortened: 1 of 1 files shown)


Debilski:
Now something common like `bot.say(f"We hate you {bot.enemy.team_name}")` will trigger a failure with the wrong enemy. Not really a fan of this hard restriction.

otizonaizit:
Can relax it a bit, I don't know, 30 chars. Your example already looks 
bad if the event team has a very long name. Or we force ellipsis if it's 
too long

11 Jun 2025 17:44:41 Rike-Benjamin Schuppner ***@***.***>:

otizonaizit:
@Debilski : do you like this one more?

Debilski:
image

Issue #895, opened 2025-06-11
Should Pelita fail when it cannot place enough food pellets?


Debilski:
```

❯ while pelita --size small --food 10:25 --null ; end
Replay this game with --seed 2647700704147470782
ᗧ blue team: 'Smart Eating Players'
ᗧ red team: 'Smart Random Players'
Finished after 271 rounds. 'Smart Random Players' won over 'Smart Eating Players'. (225:43)
Replay this game with --seed 1762835341784623474
ᗧ blue team: 'Smart Eating Players'
ᗧ red team: 'Smart Random Players'
Finished after 59 rounds. 'Smart Random Players' won over 'Smart Eating Players'. (42:25)
Replay this game with --seed 7515036359460345987
ᗧ blue team: 'Smart Eating Players'
ᗧ red team: 'Smart Random Players'
Finished after 78 rounds. 'Smart Random Players' won over 'Smart Eating Players'. (59:25)
Traceback (most recent call last):
  File "/tmp/pelita/venv/bin/pelita", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/tmp/pelita/pelita/scripts/pelita_main.py", line 444, in main
    layout_dict = pelita.maze_generator.generate_maze(trapped_food=trapped_food,
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/pelita/pelita/maze_generator.py", line 320, in generate_maze
    left_food = distribute_food(free_tiles, chamber_tiles, trapped_food, total_food, rng=rng)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/pelita/pelita/maze_generator.py", line 96, in distribute_food
    raise ValueError(
ValueError: number of total food (25) exceeds available tiles in maze (24)

```

I assume with our default of 10 food pellets each on a *small* maze, there should not be any issue, but in principle I think that Pelita should not crash here. (Unless the number exceeds (h-2)*(w//2-2).) Or do we know from maze mathematics how many pellets are going to fit guaranteed? Then using more pellets is personal risk of course.

Debilski:
I am more worried that it is a random failure.

[...]

PR #897, opened 2025-06-12
Retry maze generation if we fail to distribute food using default values for size and food

otizonaizit:
Fixes: #895


otizonaizit:
Do we want this or sound I close the PR unmerged?

Issue #891, opened 2025-06-11
when bot respawns on an enemy in its own homezone, the kill is recorded in the next turn...


otizonaizit:
I am not sure this is the way we want stuff to work.

We have two teams and a layout:

  - `stopping_bot.py`:
```python
TEAM_NAME = "Stopping Bots"

def move(bot, state):
    return bot.position
```

  - `up_bot.py`:
```python
TEAM_NAME = "Up Bots"

def move(bot, state):
    if bot.turn == 1:
        return bot.position
    else:
        return (bot.position[0], bot.position[1]-1)
```

  - `layout`:
```
########
# .   y#
#     b#
#a     #
#x   . #
########
```

rendered:

![Image](https://github.com/user-attachments/assets/55e2a918-4d11-4393-938f-ce5c877c0f18)

If we play, this is the first turn (stopping bot `a` does not move:

![Image](https://github.com/user-attachments/assets/56f5b60f-cf6f-4e64-a70d-819569840501)

Second round, `x` moves up and gets killed (look at the kill count: the blue team gets a kill):

![Image](https://github.com/user-attachments/assets/5779968d-ffab-49c2-b5bf-ad8e40b39660)

Third round, `b` does not move, but because `x` was respawned  on top of `b`, `b` gets killed and the kill counter for the red team gets updated:

![Image](https://github.com/user-attachments/assets/ece50346-7e84-46c3-856c-b88153230ac3)

But what if we change `stopping_bot.py` to be this:
```python
TEAM_NAME = "Stopping Bots"

def move(bot, state):
    if bot.turn == 0:
        return bot.position
    else:
        return [bot.position[0], bot.position[1]+1]
```

Then the first two rounds are the same, but the third looks like this:

![Image](https://github.com/user-attachments/assets/ba9bcbd0-c2c2-4e28-b38f-4d856c56aeb0)

We see that `b` "escapes" the kill.

This is not new, but it came up because this is the reason of a subtle bug in `pelita_template/demo05_basic_defender.py` https://github.com/ASPP/pelita_template/blob/d6eed14cb833aef5c81dc3668f27ae7850ddf267/demo05_basic_defender.py :

https://github.com/ASPP/pelita_template/blob/d6eed14cb833aef5c81dc3668f27ae7850ddf267/demo05_basic_defender.py#L37

We assume in that line that the bot can never be on top of an enemy bot, so that the path to the enemy bot is always of at least length 2.

Debilski:
[...]

PR #900, opened 2025-07-16
deal with cascade kill in one step

JonnyBoi-13:
more comments are coming

Fixes: #891

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -838,6 +838,7 @@ def check_kill_death(team, turn, boundary, game_state):
             _logger.info(f"Bot {enemy_idx} reappears at {bots[enemy_idx]}.")
             # we update for the killed bot as well, seeing as they could land on an enemy
             kd_state = check_kill_death(1-team, enemy_idx, boundary, kd_state)
+    
     return kd_state
 
 def apply_move(gamestate, bot_position):
@@ -933,7 +934,7 @@ def apply_move(gamestate, bot_position):
             # This is modifying the old game state
             score[team] = score[team] + 1
         
-    # we check if we killed or have been killed and update the state accordingly
+    # we check if we killed or have been killed and update the gamestate accordingly
     gamestate.update(check_kill_death(team, turn, boundary,gamestate))                               
         
 
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -788,6 +788,57 @@ def play_turn(game_state, allow_exceptions=False):
 
     return game_state
 
+def bot_in_homezone(team, bot_position, boundary):
+    if team == 0:
+        return bot_position[0] < boundary
+    elif team == 1:
+        return bot_position[0] >= boundary
+
+def check_kill_death(team, turn, boundary, game_state):
+    
+    kd_state = {}
+    kd_state.update(game_state)
+
+    bots = kd_state["bots"]
+    score = kd_state["score"]
+    walls = kd_state["walls"]
+    shape = kd_state["shape"]
+    kills = kd_state["kills"]
+    deaths = kd_state["deaths"]
+    bot_was_killed = kd_state["bot_was_killed"]
+
+    enemy_idx = (1, 3) if team == 0 else (0, 2)
+
+    # check if we have been eaten
+    if not bot_in_homezone(team, bots[turn], boundary):
+        
+        enemies_on_target = [idx for idx in enemy_idx if bots[idx] == bots[turn]]
+        if len(enemies_on_target) > 0:
+            _logger.info(f"Bot {turn} was eaten by bots {enemies_on_target} at {bots[turn]}.")
+            score[1 - team] = score[1 - team] + KILL_POINTS
+            init_positions = initial_positions(walls, shape)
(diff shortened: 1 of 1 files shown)


pelita/game.py
```diff
@@ -788,6 +788,58 @@ def play_turn(game_state, allow_exceptions=False):
...
+    score = kd_state["score"]
+    walls = kd_state["walls"]
+    shape = kd_state["shape"]
+    kills = kd_state["kills"]
+    deaths = kd_state["deaths"]
+    bot_was_killed = kd_state["bot_was_killed"]
+
+    enemy_idx = (1, 3) if team == 0 else (0, 2)
+
+    # check if we have been eaten
```
otizonaizit, line 812: the comment could be extended with something like this:
```python
# we do this first, because we may kill someone after being respawned, so the check
# for kills needs to happen after the check for being eaten
```

pelita/game.py
```diff
@@ -788,6 +788,58 @@ def play_turn(game_state, allow_exceptions=False):
...
+        for enemy_idx in killed_enemies:
+            _logger.info(f"Bot {turn} eats enemy bot {enemy_idx} at {bots[turn]}.")
+            score[team] = score[team] + KILL_POINTS
+            init_positions = initial_positions(walls, shape)
+            bots[enemy_idx] = init_positions[enemy_idx]
+            kills[turn] += 1
+            deaths[enemy_idx] += 1
+            bot_was_killed[enemy_idx] = True
+            _logger.info(f"Bot {enemy_idx} reappears at {bots[enemy_idx]}.")
+            # we update for the killed bot as well, seeing as they could land on an enemy
```
otizonaizit, line 839: this comment could be rephrased as follows:
```python
# We run the kill check recursively until there are no killed enemies anymore.
# This is necessary to deal with "cascade" kill situations like the one described
# in GitHub issue #891 and the corresponding test cases in test_game.py 
```

otizonaizit (changes requested):
Good work, thanks a lot! I left a couple of comments.
Also, please fix the corresponding tests.

pelita/game.py
```diff
@@ -870,43 +921,22 @@ def apply_move(gamestate, bot_position):
...
-        bot_in_homezone = bot_position[0] < boundary
-    elif team == 1:
-        bot_in_homezone = bot_position[0] >= boundary
+    # bot in homezone needs to be a function 
+    # because a bot position can change multiple times in a turn
+    # example: bot is killed and respawns on top of an enemy
+    
     # update food list
-    if not bot_in_homezone:
+    if not bot_in_homezone(team, bot_position, boundary):
```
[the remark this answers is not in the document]
JonnyBoi-13: It is used twice (line 930, line 948), I would prefer to keep it.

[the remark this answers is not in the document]
otizonaizit: +1

otizonaizit:
Thanks! Great work. I like this solution: it removes an ambiguity that we had all along. Can be merged for me. @Debilski ?

pelita/game.py
Debilski: The comment confuses me

Debilski: To clarify things `# check that we have been eaten` should be `# check that we ran into an enemy`

pelita/game.py
```diff
@@ -788,6 +788,56 @@ def play_turn(game_state, allow_exceptions=False):
 
     return game_state
 
+def bot_in_homezone(team, bot_position, boundary):
+    if team == 0:
+        return bot_position[0] < boundary
+    elif team == 1:
+        return bot_position[0] >= boundary
+
+def check_kill_death(team, turn, boundary, game_state):
```
Debilski, line 797: Same here; not really a fan of exposing the somewhat ill-defined boundary (is it on the last blue column or the first red column?). If we pass the game_state, we already have all we need.

pelita/game.py
Debilski: With the auto-cascading/recursive logic, I think we don’t need the distinction anymore. We are running the algorithm for as long as it takes so the order isn’t really important.

I think we can simplify the whole structure without recursion: Starting from the current bot turn, we add its location to a list of kill locations over that we iterate (list.pop): If two or more bots share a location, respawn them and add the respawn location to the list. Exit when the list is empty.

Debilski:
> Thanks! Great work. I like this solution: it removes an ambiguity that we had all along

I stand by that it was not an ambiguity. It was well-defined and easily explained to anyone who asked (I regularly explained it to teams at ASPP and it was always pretty clear). We are fixing it for other reasons. :)

Debilski:
I tried to simplify the recursion and the case-handling/code duplication in https://github.com/ASPP/pelita/commit/b03d61ca08c56ee0e2c858d3db511a7e2a3a21d2 
If you agree, I’d like to add it to the PR.

otizonaizit:
Thank you again @JonnyBoi-13  for coming up with the original idea on how to fix this and for writing the first implementation. Thanks to @Debilski for getting rid of recursion (my personal enemy in programming) and for cleaning up the whole thing.

PR #902, opened 2025-07-18
NF: Allow for multiline speech when {} is present in speaker

Debilski:
Use case: On macOS `flite` cannot output to the speaker, but `afplay` can. Having shell mode allows us to run two commands after another.

    speaker: ./flite/bin/flite -f {} out.wav ; afplay out.wav

Not really sure I like it. Maybe I should just add a shell script wrapper to flite that does the same thing?


otizonaizit:
I think it's ok like this, iff you add a comment in the code revision why this is needed ;)

Issue #574, opened 2019-03-21
[Discussion] More than two team colours during tournament


Debilski:
I just had this weird thought: What if we allowed for more than just red and blue colours in the UI? Especially (or perhaps only) during the tournament. It would certainly make it easier to find one’s team on the map when each group has a colour they are supposed to identify with.

In 2018 we already used 5 different post-it colours to indicate to which group a laptop would belong to and when we presented the repo statistics (which team had how many commits and at what hour), I tried re-using the same colours in these graphs to make the connection more obvious.

I guess the food could still be red and blue, we’d just colorise the bots differently during tournament (if we find enough good looking colours or colour combinations).

Any thoughts on this?

jni:
As long as the palette is colourblind friendly...

Anyway, I'm neither in favour nor opposed to this. It is fun but it does add another point of failure (how do you control the team/colour matching, if it's the team's responsibility how do you ensure they're all different... etc.)

otizonaizit:
it seems to me this is a nice-to-have but certainly in the back of our priority list with pelita. And, if we ever go for it, then food should be colored accordingly, so that each game is consistent and understandable on its own outside of the tournament. In order to find five colors we could of course get 5 equally spaced colors on the viridis color map to be on the safe side, no? By the way @jni, are red and blue color-blind friendly enough?

Debilski:
With regards to the colours: there could be variants that draw a border around the bots as well (needs to be projector friendly: we cannot do yellow on a white background) and therefore we’d basically have two colours for each bot (that could be the same, if a border is not needed). If get really crazy, we can also do striped versions.

For the tournament, I would fix the colourscheme in the tournament config (either free-form or with a pre-defined palette of good colours). The users have no say in it.

Sure, it is very low priority, but I didn’t want to forget about it. :)

Debilski:
Tested on https://www.color-blindness.com/coblis-color-blindness-simulator/ , if its not Monochromacy/Achromatopsia, then the blue at least stands out a little. But I don’t really have experience in interpreting this data. (I think we would have gotten complaints though, if it were too bad.)

jni:
As I understand it red/blue is ok contrast, and orange/blue is the best contrast you can get for most people. For many colours we can use the seaborn colorblind palette:

https://seaborn.pydata.org/tutorial/color_palettes.html

You can do a lot better than viridis if you don't have the restriction that colours have to be sequential...

PR #903, opened 2025-07-18
Adding publish mode and handling to pelita tournament

Debilski:
Still a draft. Will attempt to merge once the frontend is in a usable (= usable by other people than myself) state.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -414,6 +414,8 @@ def setup_game(team_specs, *, layout_dict, max_rounds=300, rng=None,
         #: Name of the teams. Tuple of str
         team_names=team_names,
 
+        team_specs=team_specs,
+
         #: Additional team info. Tuple of str|None
         team_infos=team_infos,
 
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -17,7 +17,7 @@ from .base_utils import default_rng
 from .exceptions import NoFoodWarning, PelitaBotError, PelitaIllegalGameState
 from .gamestate_filters import noiser, relocate_expired_food, update_food_age, in_homezone
 from .layout import get_legal_positions, initial_positions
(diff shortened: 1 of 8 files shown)


Debilski:
Ready for a first review (in conjunction with https://github.com/ASPP/pelita-webtournament ).

tl;dr

    docker run -p 3000:3000 debilski/pelita-webtournament

Open in a browser: http://localhost:3000/

Go to: Live tournament → Single game

    run pelita --http-post http://localhost:3000/api/collect --stop-at 0

Works over the network but obviously profits from a fast connection. Multiple computers can watch at the same time. Few minor graphic bugs are known to the author. More explanation to follow :)

Debilski:
Closes #574.

otizonaizit:
Ouch, I tried to review the changes, but I am a bit overwhelmed... any chance that this could be either split out as the ci_engine, so that is almost does not affect pelita main, or otherwise I think I would need to be hand held in understanding what is happening and some docs will have to be added somewhere...

Debilski:
Ah, yes, I had already started a longer document with the reasoning behind choosing httpx here (and why I chose it over zmq) but I forgot to finish it.
But then again, in this PR there are basically only two changes. One is adding some metadata object in the tournament that can be serialised to a viewer; the other is just a new viewer class `POSTPublisher` to have Pelita make POST requests with the current game state to some URL. Not sure if there is actually much to split out here.

otizonaizit:
So maybe you could add a little comment in the code to explain why httpx? And why it is imported within the class and not at the module level? Also maybe put in the docs of the new class an explanation of its purpose (to publish on th etournament website or whatever)? 
I am thinking about us looking back at this in 3 years and wondering what this is all about.

Debilski:
I’ll postpone the answer (and merge) slightly: I’ve just had another idea come up that I wanted to test first before the write up.

PR #905, opened 2025-07-19
Increase bot shadow distance to 2 and darken the shadow in the GUI

otizonaizit:
As discussed after the tournament at the TU in summer 2025. Now attack and defense are more balanced, and the presence of tunnels and chambers is increasing the chance of killing bots. The food shadow of 1 is just a little too small and makes it a bit too easy to block big portions of food. By increasing it to 2 we are going in the middle between 1 and 3, which is the number with which we started the food shadow system. Let's see.


Debilski:
A bit close to the colour of the cell selection maybe?

image

otizonaizit:
> A bit close to the colour of the cell selection maybe?

What do you mean? That it is too close? Yes it is, so maybe we could make the two colors identical instead of slightly different. But I don't think it's a problem either way.

Debilski:
Additional bug I found: Shade should not cross the middle line.

Debilski:
Yes yes. I just hope you choose the slightly non-greyscale colour (#CDC7C2) for its aesthetic and not because you measured it against your reddish screen in night mode. :)

Issue #82, opened 2012-05-09
UI Viewer should include browseable history


Debilski:
A browseable history would make it easier to revisit past events which might be useful in some situations. (A bit further down that path would be an import–export function.)

Debilski:
I was right. I wasn’t going to happen before Kiel. ;)

PR #912, opened 2025-08-23
Add History to Tk Viewer

jbdyn:
As the title says.

You can now go backward and forward in time.
Also, the button styles were changed.

Fixes: #82

### DIFF
--- a/pelita/ui/tk_canvas.py
+++ b/pelita/ui/tk_canvas.py
@@ -355,7 +355,7 @@ class TkApplication:
         # previous
         tkinter.Button(self.ui_status_00,
                        text="previous",
-                       command=self.show_previous,
+                       command=self.button_show_previous,
                        **BUTTON_STYLE).pack(side=tkinter.LEFT, expand=True, **BUTTON_PADDING)
 
         # play/pause
@@ -367,7 +367,7 @@ class TkApplication:
         # next
         tkinter.Button(self.ui_status_00,
                        text="next",
-                       command=self.show_next,
+                       command=self.button_show_next,
                        **BUTTON_STYLE).pack(side=tkinter.LEFT, expand=True, **BUTTON_PADDING)
 
         tkinter.Button(self.ui_status_01,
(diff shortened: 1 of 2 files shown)


otizonaizit:
@jbdyn : maybe instead of unicode characters for the navigation controls we could use images.

As the story about using custom fonts in Tk is very, but very very very sad, I think this is the only reasonable option left?

Looking at [this](https://tkdocs.com/tutorial/fonts.html#images) and at [this](https://tkdocs.com/tutorial/widgets.html#button) it seems it should be straightforward to load an image with `pillow` and stick it into the button instead of the text label. What do you think?

otizonaizit:
Great, this solves the problem then!

> BUT: These images have to be rasterized files, so fixed in size (now PNG, 32x32 px) 🥶 . This could be a problem on HiDPI displays; have not tested it. We might need to open the images in a `PIL.Image` object, then resize it according to the dpi of the screen and then pass that to the `tkinter.PhotoImage` class. Or use `PIL.Imagetk.PhotoImage` (or similar) 🤷

@Debilski was experimenting with scalable fonts+canvas in #910, so maybe we can revisit the icon scaling once those experiments show the way to go? But this is not a blocker for this PR in my opinion.

> And I labelled the quit button with `QUIT` again as I could not figure out a good icon for it. The power symbol we had before is like an "on / off", which is not really on point as it just turns "off", and an "X" just looked odd down there. 😅

Ah, OK. If I remember correctly @Debilski had the same comment some time ago on the use of that symbol to mean "switch off"?

I personally don't think that "on/off" is ambiguous: it is immediately clear to me what the button is supposed to do, and in the worst case you understand and memorize the first time you hit it by thinking that it will start a new game? :-))

But I am not attached to having an icon for it, we can leave it with "QUIT". In my comment https://github.com/ASPP/pelita/pull/882#issuecomment-2907830772 on #882 I was even arguing to remove the button completely (and just a couple of months later I was arguing to keep it at any cost while talking  with you on the train...). By the way, the idea of using the symbols for navigation was discussed already in #882. I am happy to see it realized now :-)

jbdyn:
> Should we maybe rebase the styling things into a new PR?

That is a good idea, as icons seem to come with new problems (so unfortunate ...).

I would revert to simple labels `previous`, `play/pause`, `next`, `slower` and `faster` for now and the history functionality can be merged.

In a dedicated PR about the button looks, we can do

- icons,
- icon sizes,
- button styles,
- and probably also button arrangement etc.

without the need to hurry.

Sounds good?

jbdyn:
Okay, I rewrote history a bit and the UI looks like this now:

buttons

Functionally nothing changed, of course. Should be ready to merge now.

PR #916, opened 2025-09-03
Fix send_exit behaviour

Debilski:
Fixes a bug (non-existent instance variable) introduced in #890 that caused longer timeouts and potential hangs during exit (including when `__del__` is called on a game state).


Debilski:
(Experimenting a little to find more race conditions; will take care of merging myself later)

Issue #914, opened 2025-08-28
UX improvements for the history viewer


Debilski:
Some stuff I noticed when using the history viewer (open for discussion):

I’d put the `<` and `>` (little unfortunate that `<` and `>` are the keys for slower and faster but we’ll have to live with this) keys next to each other, so

    backwards  forwards  play/pause

One typically wants to jump between backwards and forwards so there is less mouse travel.

When in history mode, it maybe should be written somewhere that this is history mode. Or the colour of the background or whatever should change.

When in history mode, hitting the play/pause button should bring you back to the most recent step (and exit history mode). Only on a second press should the game run (same goes for hitting the enter key for next step). I think before making a completely new step you want to see where you’re coming from, otherwise it can be a little confusing. (play/pause could be renamed to ‘jump back’ when in history mode so this would make things clear).

otizonaizit:
> I’d put the `<` and `>` (little unfortunate that `<` and `>` are the keys for slower and faster but we’ll have to live with this) keys next to each other, so
> 
> ```
> backwards  forwards  play/pause
> ```
> 
> One typically wants to jump between backwards and forwards so there is less mouse travel.

I agree.

> When in history mode, it maybe should be written somewhere that this is history mode. Or the colour of the background or whatever should change.

Yes, good idea. Maybe not something so intrusive as changing the background. But a red dot near the text "viewing history" or so in the button bar somewhere maybe?

> When in history mode, hitting the play/pause button should bring you back to the most recent step (and exit history mode). Only on a second press should the game run (same goes for hitting the enter key for next step). I think before making a completely new step you want to see where you’re coming from, otherwise it can be a little confusing. (play/pause could be renamed to ‘jump back’ when in history mode so this would make things clear).

I do not agree so much with this. I think that if we implement the above suggestion of signaling explicitly when you are in history mode, I think it is clear enough what is happening without the need to change the ergonomics and/or the button names of the UX.

Debilski:
> > When in history mode, it maybe should be written somewhere that this is history mode. Or the colour of the background or whatever should change.
> 
> Yes, good idea. Maybe not something so intrusive as changing the background. But a red dot near the text "viewing history" or so in the button bar somewhere maybe?

Yeah, something like this would also work.

[...]

PR #918, opened 2025-09-09
switch history button position and playback mode status in the GUI

otizonaizit:
Additionally to the changes visible in the screenshot, the left and right arrow keys have been added as keyboard shortcuts for `previous`/`next`.

screenshot1

Fixes: #914


Debilski:
I might have removed the text on the current step already (instead of only on the new step when moving backwards and forwards) but I guess that’s up for interpretation.

PR #920, opened 2025-09-10
RF: Simplify the history status

Debilski:
I think this makes the code a little nicer :) (Plus it fixes a bug: SHIFT+ENTER would flicker the status)

Arguable change: When we return to the current step we exit ‘playback mode’ again. (I like this behaviour more but most of all it was way easier to implement.)

### DIFF
--- a/pelita/ui/tk_canvas.py
+++ b/pelita/ui/tk_canvas.py
@@ -904,7 +904,8 @@ class TkApplication:
             for idx, label in enumerate(self.ui_bot_indexes):
                 label.configure(fg=bot_colors[idx] if turn == idx else "#bbb")
 
-
+        # update history status bar
+        self.set_history_mode_status(self.get_next_pointer() in self.history)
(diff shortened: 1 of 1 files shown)


otizonaizit:
> I think this makes the code a little nicer :) (Plus it fixes a bug: SHIFT+ENTER would flicker the status)

ah, ah, ah, well had I remembered the existence of `draw_status_info` I may have come up with a similar solution :-))

I personally don't care about the `playback mode` being exited at the current step or only  when asking for the next. If `playback mode` means that whatever print statements you have in your code won't be shown, then I think being in `pause` at the current step means that we are still in `playback`. But really, given that this implementation is so much easier, I'd stay with the behavior implemented in this PR.

Merging.

Debilski:
For me it is about the anticipation. When the ‘playback mode’ sign vanishes, I know that the *next* move will be a new one.

Issue #877, opened 2025-04-30
make layout overlays available from Bot


otizonaizit:
After #860 it is possible to overlay a color on top of specified tiles in the maze in the GUI.

This should be available directly from the `Bot` object. I imagine an API like this:

```python
bot.overlay({(x0, y0) : '#AABBCC', (x1, y1): '#BBDDEE'})
```

I guess this should bne treated the same way we treat `bot.say`.

Debilski:
How do you tell the bot the URI of the viewer?

otizonaizit:
Wait, for `bot.say` you don't need any address, no? So, just the same for `bot.overlay`.

Debilski:
Yeah but then it is a completely different concept to how the overlay is done now, if we want to overlay graphics on the main viewer. (Will also need extra switches to turn this off/on then.)

otizonaizit:
Oops, then it will take a bit longer to implement...

otizonaizit:
Why are you worried about users overusing the feature? The API with `can_do_overlay` and `set_overlay` sounds complicated, and I am not sure what is it useful for. 

I think the overlay should be shown by default when using the "debug" button in the GUI. This could and maybe should become a CLI switch `--debug` .

The client would then send the overlays only when debug is on, and otherwise silently drop them.

otizonaizit:
I don't think the timeout is a problem. We have already `--no-timeout` and people are using it already when they debug using the debugger. So, if their overlays take long, they'll resort to run with `--no-timeout`. 

>The two CLI switches are because you typically are only interested in the overlays of your own bot. I don’t think it makes a lot of sense to switch back and forth between the blue and the red team’s overlays in UI.
>Thinking about this: As a user, I probably would not want that an enemy team can see my overlays. So how would we ensure this? (Maybe make it impossible for server games?)

But overlays are a strictly debug feature, so if you use it against locally available bots or in network games. The network clients are not your "enemies" in the current tournament ever. Why is there any need to protect their overlays? Also, if we were to allow playing against the current enemies, it is enough to say that overlays are visible to the enemy, like your strategy is visible to the enemy after all.

I think this is all over-complicating the thing. Let's start with a basic implementation that just shows the overlays when in debug mode, and fix problems when they appear.

PR #934, opened 2025-10-05
Add `paint_background` method to `Bot` object to paint cells in the maze with user-defined colors and show it in TK viewer in debug mode

otizonaizit:
Expected usage would be a bot like this:
```python
TEAM_NAME = "overlay"

def dist(pos1, pos2):
    return abs(pos1[0] - pos2[0]) + abs(pos1[1] - pos2[1])

def danger_zone(enemy, size):
    x, y = enemy.position
    return [ (i,j) for i in range(x-size, x+size+1)
                   for j in range(y-size, y+size+1)
                   if dist((i, j), (x, y)) <= size ]

def move(bot, state):
    danger0 = danger_zone(bot.enemy[0], size=2)
    # paint green danger zone aroung enemy0
    bot.overlay([(coord , '#FFFFA8') for coord in danger0])
    danger1 = danger_zone(bot.enemy[1], size=2)
    # paint yellow danger zone around enemy1
    bot.overlay([(coord , '#96FF96') for coord in danger1])
    return bot.position
```

which paints a danger zone around the (noised) enemies when running the TK viewer in debug mode:

screenshot1

@Debilski : I can not understand why the overlays are only deleted at the end of the round instead of being deleted at every turn. I verified that the `game_state` sent to TK is the correct one, i.e. the overlays are only there when the bot send them. It is a minor problem though, so even if we don't fix it, I think this PR should go in. I will then copy the code above in a demo bot in `pelita_template` and add it to the docs there.

@Debilski: This PR also makes #860 obsolete? Because #860 is adding a lot of complexity to the code, maybe we could get rid of it if this PR is merged?

Fixes: #877

### DIFF
--- a/test/test_team.py
+++ b/test/test_team.py
@@ -6,7 +6,7 @@ from pelita.game import SHADOW_DISTANCE, play_turn, run_game, setup_game
 from pelita.gamestate_filters import manhattan_dist
 from pelita.layout import initial_positions, parse_layout
 from pelita.maze_generator import generate_maze
-
+from pelita.exceptions import PelitaBotError
 
 def stopping(bot, state):
     return bot.position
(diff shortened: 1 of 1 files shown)


otizonaizit:
by the way, I don't agree to make Ruff checks fatal when checking PRs...

Debilski:
Ok, all in all it looks easier than I had thought. :)


> @Debilski : I can not understand why the overlays are only deleted at the end of the round instead of being deleted at every turn. I verified that the `game_state` sent to TK is the correct one, i.e. the overlays are only there when the bot send them.

You have to explicitly reset the overlay values to `[]` for the indexes of the non-current bots in the gamestate.


> 
> @Debilski: This PR also makes #860 obsolete? Because #860 is adding a lot of complexity to the code, maybe we could get rid of it if this PR is merged?

Hmm, not sure I understand. What parts of #860 would be obsolete/are complex? I don’t see much in there that would be superfluous now.

---

> ```
> def move(bot, state):
>     danger0 = danger_zone(bot.enemy[0], size=2)
>     # paint green danger zone aroung enemy0
>     bot.overlay([(coord , '#FFFFA8') for coord in danger0])
>     danger1 = danger_zone(bot.enemy[1], size=2)
>     # paint yellow danger zone around enemy1
>     bot.overlay([(coord , '#96FF96') for coord in danger1])
>     return bot.position
> ```

I think having take `bot.overlay` a list of `(pos, colour)` *and* making it automatically append in the background is the worst of both worlds. We should either rely on the user creating the full overlay list before calling `bot.overlay` (and multiple calls of `bot.overlay` will overwrite previous values); *or* we should make it single-coordinate only and have the more explicit API of `bot.overlay(pos, colour)`.

That said, why not – for now – stick with the API that is already there?

    overlays = [
        {'fill': ''#FFFFA8', 'pos': danger0},
        {'fill': ''#96FF96', 'pos': danger1},
    ]
    bot.overlay(overlays)

We can still argue about the exact syntax of the dicts. The reason why I prefer this is that `"fill"` is not the only possible overlay type, it is just currently the only one implemented. I think it definitely makes sense to add one or more ‘number’ overlays – either in displaying a single number in the centre (or in multiple cell corners) or by displaying a ‘percentage bar’. Something like this to show actual values for a cell and not just a qualitative colour:

image

In any case, I don’t think we will stop after having added background fill.

otizonaizit:
> You have to explicitly reset the overlay values to `[]` for the indexes of the non-current bots in the gamestate.

Yep, right, I'll fix it.
 
> > @Debilski: This PR also makes #860 obsolete? Because #860 is adding a lot of complexity to the code, maybe we could get rid of it if this PR is merged?
> 
> Hmm, not sure I understand. What parts of #860 would be obsolete/are complex? I don’t see much in there that would be superfluous now.

Well, the whole "standalone" mode could go. It's a lot of little things sprinkled all over the code base.

> I think having take `bot.overlay` a list of `(pos, colour)` _and_ making it automatically append in the background is the worst of both worlds.

Well, the "worst of both worlds" sounds a bit extreme, but thanks for the discussion. It helped me re-think this.

So, I think the important bit is that we can make the data structure as complicated and nested as we need it to work nicely with it internally and to be easily extensible, but that complexity does not need to be exposed to the user. The original list of dictionaries is hard to get right at first sight. We fought with it a lot when @jbdyn and me were trying to improve maze generation. So, what about something like this:

- internally we use a list of dicts: this is the most flexible structure that allows us in the future to add new things without having to re-think the data structure every time
- To the user we expose methods with telling names and the simplest possible API, which in general take one single position and one single property has arguments

So, for now we could have `bot.set_pos_background(position, color)`, where, for simplicity we could even have a default color (light yellow or so), so that if all they want is to highlight a region of the maze to verify that their code is correct, like I do with the "danger zone" example above, then they can just do:

```python
def move(bot, state):
    for pos in danger_zone(bot.enemy[0]):
        bot.set_pos_background(pos)
```

If we decide to implement overlays with text, then the method will be `bot.set_pos_text(pos, text="")` or something like that. The implementation in `Bot` will use the private `_overlay` attribute to store the list of dictionaries:

```python
[
{'fill': '#FFFFA8', 'pos': (x0, y0)},
{'fill': '#AABBCC', 'pos' : (x1, y1)},
...
]
```
and at every call of `bot.set_pos_something` update the dictionary.

If we were to add something more in the future, then the new thing will just be a new key in the dictionary.

This way usage is pretty straightforward, and we still can had all fancy things you may want in the future.

What do you think? I can update this PR with this new design and we can see how it looks.

Debilski:
Yeah, paint will do for now.

test/test_team.py
```diff
@@ -767,3 +767,96 @@ def speaking_team(bot, state):
...
+    parsed = parse_layout(test_layout)
+
+    def overlay_team(bot, state):
+        for pos, color in overlay:
+            bot.paint_background(pos, color=color)
+        return bot.position
+
+    state = setup_game([overlay_team, overlay_team], max_rounds=2, layout_dict=parsed, raise_bot_exceptions=True)
+
+    overlay_in_maze = [{'pos':coord, 'color':color} for coord, color in overlay if coord[0] < 32]
```
Debilski, line 798: Idk, but for simplicity I guess we could just write it out for the two entries? :)

test/test_team.py
```diff
@@ -767,3 +767,96 @@ def speaking_team(bot, state):
...
+    for idx in range(4):
+        state = play_turn(state)
+        gs_overlay = state['overlays'][idx]
```
Debilski, line 801: I think what could be added is a test that shows that different bots can have different overlays? And that they are deleted after a turn.

PR #936, opened 2025-10-12
add `--debug` flag to the pelita command to start the TKviewer in debug mode

otizonaizit:
This allows to start pelita with:

```bash
pelita --debug my_player.py my_other_player.py
```

and have the TK-viewer automatically start in debug mode. What students currently do is to start pelita and then try to hit the debug button as soon as possible.

Further possible improvement that would require some thought: when passed the `--debug` flag also pass `--stop-at 0` so that the game does not start right away. This may be closer to what a user want when they directly start in debug mode.

### DIFF
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -349,14 +349,19 @@ def main():
     elif args.viewer == 'tk':
         geometry = args.geometry
         delay = int(1000./args.fps)
+        debug = args.debug
         stop_at = args.stop_at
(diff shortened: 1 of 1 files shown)


otizonaizit:
> I also agree on making it stop immediately when `--debug` is passed. Starting the real game is as easy as hitting the space key, so this should not be a big deal.

I implemented this in https://github.com/ASPP/pelita/pull/936/commits/8308a9547caf5b051358d1b417dcbd8596642b81 . If you like it I would merge.

PR #952, opened 2026-03-30
Do not raise in main()

Debilski:
There is not really a use in catching these errors or see the stack trace so we should default to `parser.error`.


otizonaizit (approved):
LGTM, thanks!

PR #955, opened 2026-04-14
RF: Make small layouts bigger and big layouts smaller.

### LINKED ISSUE #954: Change default maze sizes
I think both small (16x8) and big (64x32) maze sizes are not really useful (other than maybe for explicit stress testing or UI design – a small maze has larger UI elements and bots become a pacman faster).

Proposition: small becomes 24x12 and big becomes 48x24 (with some intermediate food spec). We can keep the other sizes under the names tiny and huge (we can also delete them but right now both maze sizes already have a food spec attached so we would throw that away also).

Also, we should allow `--size 12` and it should just give us a maze `--size 24x12` because it just makes sense.

Debilski:
Closes #954

New sizes: 

         'tiny': (16, 8),
         'small': (24, 12),
         'normal': (32, 16),
         'big': (48, 24),
         'huge': (64, 32),

I didn’t think much about the food specification. Any ideas? Technically this should scale with the area. I chose 50 now for the new big because it is a nice round number.

Also, specifying only the *height* will create a maze with twice the width.

### DIFF
--- a/pelita/game.py
+++ b/pelita/game.py
@@ -54,16 +54,20 @@ INITIAL_TIMEOUT_SECS = 6
 
 #: Default maze sizes
 MSIZE = {
-         'small' : (16, 8),
+         'tiny': (16, 8),
+         'small': (24, 12),
          'normal': (32, 16),
-         'big'   : (64, 32),
+         'big': (48, 24),
+         'huge': (64, 32),
         }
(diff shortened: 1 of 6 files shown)


otizonaizit (approved):
LGTM, thanks!

PR #956, opened 2026-06-02
Get rid of \x04 character in JSONL format and trust json.dump

Debilski:
The `\x04` (EOT) character was introduced by us to separate JSON records, but when a JSON record is properly exported without indents, it is enough to separate them with `\n`. In fact, all external tooling (`jq`) expects a plain `\n` in the JSONL format.

Semi-breaking change: This breaks historic replay files, but then again, historic replay files only work with some manual restructuring anyway (and now we can at least use `jq` to do the restructuring).

### DIFF
--- a/pelita/scripts/pelita_main.py
+++ b/pelita/scripts/pelita_main.py
@@ -365,7 +365,7 @@ def main():
         if pelita.game.controller_await(viewer_state, await_action='set_initial'):
             sys.exit(0)
 
-        old_game = Path(args.replayfile).read_text().split("\x04")
+        old_game = Path(args.replayfile).read_text().split("\n")
         for state in old_game:
             if not state.strip():
                 continue
--- a/pelita/viewer.py
+++ b/pelita/viewer.py
@@ -165,18 +165,18 @@ class ReplyToViewer:
 
 
 class ReplayWriter:
-    """ A viewer which dumps to a given stream.
+    """ A viewer which writes JSONL to a given stream.
     """
     def __init__(self, stream):
         self.stream = stream
 
     def _send(self, message):
-        as_json = json.dumps(message, cls=SetEncoder)
-        self.stream.write(as_json)
-        # We use 0x04 (EOT) as a separator between the events.
-        # The additional newline is for improved readability
-        # and should be ignored by the Python json reader.
-        self.stream.write("\x04\n")
+        # Write a record with minimal spacing
+        # indent=None should ensure that no new lines are introduced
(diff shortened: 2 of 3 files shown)


otizonaizit:
is this breaking the replay of the old tournaments you have published here https://aspp.github.io/pelita-webtournament/tournament-history ?

Debilski:
The web viewer filters out `\x04` anyway because it is non-standard so it should work with both variants. The Python viewers will not be able to read the old files anymore but you can just `sed` away the `\x04` character and it should be fine. (Depending on how old the replay is, there will be a need for deeper rewrites, however.)

Idk. I can also make it such that Pelita filters out superfluous `\x04` characters so that old replays still work but I think we should just rewrite the replay files to the new format.

All that said, we should eventually create a new replay format: These files regularly take up between 2 and 20 MB, which is almost the same size as a *video* from the game would have. Removing repeated values gives a shrink factor of 10% and lower (which is still too high when storing 1000s of games for CI purposes) but this should be made official. And then, obviously, the format should include a version string. :)

otizonaizit:
OK, LGTM.

Regarding the replay format, how big are the biggest replay files after compression via zlib or something similar?

Debilski:
> how big are the biggest replay files after compression via zlib  

~ 90 kB (only marginally bigger than when filtering out duplicate data manually before zipping btw)

The naive optimum (without food relocation, completely deterministic), just encoding the direction a bot takes, would be something like

Maze = 16 * 32 bits (or 16 * 16 since rotated but let’s ignore this) = 64 Bytes (without the border it would be 53 bytes)
Initial food = 64 bytes
Bot moves: 1200 steps * log_2(5) ⪅ 2787 bits ⪅ 385 bytes

Adding a bit of metadata, this would be maybe 1kB. Of course food relocation complicates things. Does this calculation help in any way? Probably not. :)

otizonaizit:
well, you said that we need a different format because the replay files are too big. If after compression they become 90kB each, I would say we don't need to bother.

Debilski:
Yeah I guess that is the way to go then. I would still move the repeated values to an initial configuration dict eventually. For reference a few more numbers (compression with zstd):

```
Random bots (no speak, only little food relocation): 3538411
Compressed                                             57699
Compressed (without food_age)                          46270
Compressed (+ without say)                             46095
Compressed (+ without team_time)                       30037
Compressed (+ without requested_moves)                 22037
```

Thinking about it `team_time` is remarkably costly, producing 8 bytes (1 float) at each step with only marginal value. I guess this can be at least rounded off for long term and CI storage.

PR #958, opened 2026-06-17
rename `Bot.error_count` to `Bot.timeouts`

otizonaizit:
I did not touch the CI engine, because I am not sure which instances of the name `error_count` and/or `errorcount` etc. are actually relevant for this PR. The tests are passing, but something else may be broken in the CI because of this?

I will create a companion PR on pelita_template to rename the attribute in the documentation


Debilski:
I actually made the same change already for CI and the viewers (https://github.com/ASPP/pelita/commit/bd24133) but I called it `num_timeouts` to make it less confusing with `game_state['timeouts']` which has a different type/value signature.

otizonaizit:
hmm, I don't see that commit in `main`... I am personally not affected by the confusion with `game_state['timeouts']` : but anyway, this will be fixed once you have a final version of #851, no? So I think this could go in already and will be superseded then?

I am afraid this is a bit the problem with the several "big" PRs that are lingering... little fixes like this become harder if we have to take care of the big PRs that are not in a mergeable state for years/months...

otizonaizit:
in the TK viewer the user facing UI shows the name `Timeouts`: that is why I chose that name for the attribute to `Bot` and for consistency I used the same name for the rest of the API.

Debilski:
> hmm, I don't see that commit in `main`...

It should be there. It is merged already.

> I am personally not affected by the confusion with `game_state['timeouts']`

I was quite a bit when working on the CI engine, that’s why I changed it.

> but anyway, this will be fixed once you have a final version of #851, no? So I think this could go in already and will be superseded then? 
> I am afraid this is a bit the problem with the several "big" PRs that are lingering... little fixes like this become harder if we have to take care of the big PRs that are not in a mergeable state for years/months...

Yeah, with #851 this might change but I think until I worked on the CI engine I never thought about it. So in that sense #851 is not a blocker but a fluctuating WIP branch that we can port back new features and renamings onto. :)

otizonaizit:
so, I have made this compatible with the other changes you made. So the thing now is called `num_timeouts` everywhere, except for the name of the attribute `Bot.timeouts`m which stays this way to be a) shorter, b) compatible with what is shown in the UI.

Debilski (approved):

Issue #959, opened 2026-06-24
Discussion: get rid of timeouts and introduce a time quota


otizonaizit:
### Current state
The game master tracks the time spent by the client to return a move and when this time exceeds a configurable timeout (default 3s), it logs a non-fatal error for the client and assigns a random move instead. After a certain number (default: 5) of non-fatal errors has been logged, the client loses the game.

### Motivation for timeouts
- we want to encourage efficient code and punish inefficient code
- we want the game to still be enjoyable to watch, and if it takes more than a couple of seconds per move, the game becomes very boring

### Problems with the current state
- timeouts are basically non-actionable in client code:
  - the client code doesn't know that it is computing too long until after the fact
  - the game master assigns a random move, so the client code may be pushed in an inconsistent state, because it assumes that the move has indeed succeeded
  - if the computation lasts longer, the client code may get several timeouts in a row
  - historically I have not seen a single client dealing with timeouts in any way
- the time includes the network time, i.e. the time it takes for the move to travel over the network to the game master: for network games this is a serious issue, because a network glitch may assign a timeout to a bot that is otherwise very fast

### A proposal
- the client gets assigned a configurable "time quota" at the beginning of the game, let's say 20 seconds
- the client keeps track of the time it spends while calling the move function and subtracts it from the time quota
- if the time quota is consumed, the client raises an Exception and loses the game

### Advantages of this proposal
- the client code can monitor the used quota and choose to do heavy computations if there is still enough time
- the client can allocate the assigned time quota freely, i.e. it can decide to do some heavy calculations at the beginning
- we get rid of "non-fatal" errors in pelita, and as a result simplify the logic a lot
- by making it possible to monitor the time quota and act accordingly we are making it possible for the students to write code that takes care of time
- the time is now computed locally, so the network latency is not relevant anymore: we still need a network timeout, of course, to stop a game if a client is not responsive anymore, but this way the two concepts (time quota vs network timeout) are completely separated and not interleaved as they currently are

Note that we are still fulfilling the original motivation for the timeouts. The game may look strange if the first move takes a lot, like consuming half of the allotted time, but at least we know it won't last long.

### Disadvantages of this proposal
- the time management gets done completely on the client side, so the game master has to trust the client on this

While this is a problem in theory, I don't think it is a problem in practice. The implementation of this proposal has to make sure that you can't mess with the time quota from within the move function and this is all the security we need. Given that we don't run tournaments with self-written clients and that we control the network server, I think this is safe enough.

[...]

PR #962, opened 2026-06-30
Make first timeout fatal

otizonaizit:
This PR just changes the default `error_limit` to 1.
For a discussion see #959

The natural follow up would be to change the code to remove the concept of "non-fatal" error, remove all the code related to counting timeouts, simplify the UI to stop showing the timeout count, adapt the docs in `pelita_template`.


otizonaizit:
superseded by #964

PR #964, opened 2026-07-01
Remove the concept of non fatal errors (IOW: make timeouts fatal) 

otizonaizit:
This PR is an alternative to #962 that gets rid of the concept of non-fatal errors everywhere in pelita.

The only visible change for the user is that now the first timeout is a fatal error.

For an in-depth discussion see #959 .


otizonaizit:
@Debilski : merge?

Debilski:
Looks ok. Didn’t have time to search the code base for any oversights but we can find them later.

Maybe, we can use the occasion to rename `fatal_errors` to `errors`.

otizonaizit:
I think for now I would leave `fatal_errors` so that the intent is still clear in the name. When and if we do the big `game_state` rewrite from `dict` to a specialized class we could rename things. Also, at the moment it helps in search replace to have this non-generic name...

If you approve in GitHub I can merge as-is.

Debilski (approved):

PR #968, opened 2026-07-08
NF: Introduce food_eaten counter

### LINKED ISSUE #911: Remove number of deaths from UI (and maybe add food eaten?)
Having both kills and deaths feels superfluous to me (it was introduced in f4ec25940bd8490587ea6bd7395427652e2d641c) and removing it would make the status info smaller.

OTOH, we could add the number of pellets eaten by a team (though maybe in a second step: this needs a backend attribute). And to save space, the names could become small icons (maybe hard to make it look right).

Debilski:
Adds a new counter `gamestate["food_eaten"]` and `bot.food_eaten` that tracks the number of food pellets eaten by a specific bot.

image

If there is consensus, I would also remove the deaths from the UI as they are identical to the number of kills of the other team.

Closes https://github.com/ASPP/pelita/issues/911


otizonaizit:
Is this a team property or a bot property?
 I'm not sure what is this useful for?

Isn't this `30 - len(bot.enemy.food)`?

If it is just for display in the UI, maybe we don't need to expose it in the bot object?

In the GUI I would call it simply `food` instead of `food eaten`.

Agreed on removing `deaths`

Debilski:
It is on a per-bot level so it is a bit more accurate than `30 - len(bot.enemy.food)`. Not sure if it is needed but if we have bot.kills then we might as well include this.

otizonaizit:
> It is on a per-bot level so it is a bit more accurate than `30 - len(bot.enemy.food)`. Not sure if it is needed but if we have bot.kills then we might as well include this.

Wait, so `bot.kills` is a bot-level counter, not a team-level counter? Why did we do this?

Debilski:
https://github.com/ASPP/pelita/commit/b351160ba35e563ec21eff0151848de52deee79a 🤷

Debilski:
I guess we can also postpone the merging of this until we have decided whether we want to thin out the Bot attributes.

PR #970, opened 2026-07-08
ENH: Always return a new set in sample_nodes

Debilski:


jbdyn:
This is ready to merge, right?

